PR #25, opened 2015-01-05
Add more typographer replacement rules

svbergerem:


puzrin:
I'd like to keep typographer limited with neutral, not conflicting, and compact set of rules.

1. <3 is not typography, it's emoji, and it's conflicting without additional checks (1+1<3).
2. Not sure that arrows replace will not conflict with anything else.
3. I know that basic rules will not satisfy everyone, especially with language-dependent typography. Idea is to make it "good enougth" and allow easy plugins writing instead of collaborating everyting into core.

---

- You can just use [markdown-it-emoji](https://www.npmjs.com/package/markdown-it-emoji) with proper aliases options, without any code change.
- Or you can use [markdown-it-for-inline](https://github.com/markdown-it/markdown-it-for-inline) to quickly add your replacements.

---

If you know, why those replacements should be in core, please explain. That's discussable. I don't like to reject PRs with help, but please understand that i have reasons to be conservative about core extentions.

svbergerem:
@puzrin I needed those changes for a project and thought they could be useful for everyone, but you are right. Those additional replacements will likely be conflicting for someone. I'll close this PR.

puzrin:
@svbergerem , what do you think about syntax from PR https://github.com/markdown-it/markdown-it/issues/89 ?

Is it convenient / common? I like arrows, but don't know what people use.

svbergerem:
@puzrin For diaspora* we use the syntax which you can see in this PR. (Implemented using markdown-it-for-inline: https://github.com/diaspora/diaspora/blob/develop/app/assets/javascripts/app/helpers/text_formatter.js#L15)

I also don't really know what is a common syntax for arrows.

PR #41, opened 2015-01-31
Add a new function to specify image size. Also some tests are added to t...

tatsy:
Add a function to specify image size. For example,

```md
![image](image.png =100x200)
```

will be interpreted as

```html
<img src="image.png" alt="image" width="100" height="200">
```

I also added some tests for this function, and confirmed the tests are passed on Travis CI.

Thanks.

### DIFF
--- a/lib/helpers/parse_image_size.js
+++ b/lib/helpers/parse_image_size.js
@@ -1,0 +1,72 @@
+// Parse image size
+//
+'use strict';
+
+function parseNextNumber(str, pos, max) {
+  var code,
+    start = pos,
+    result = {
+      ok: false,
+      pos: pos,
+      value: ''
+    };
+
+  code = str.charCodeAt(pos);
+
+  while (pos < max && (code >= 0x30 /* 0 */ && code <= 0x39 /* 9 */)) {
+    code = str.charCodeAt(++pos);
+  }
+
+  result.ok = true;
+  result.pos = pos;
+  result.value = str.slice(start, pos);
+
+  return result;
+}
+
+module.exports = function parseImageSize(str, pos, max) {
+  var code,
+    result = {
+      ok: false,
+      pos: 0,
+      width: '',
+      height: ''
+    };
+
+  if (pos >= max) { return result; }
+
+  code = str.charCodeAt(pos);
+
+  if (code !== 0x3d /* = */) { return result; }
+
+  pos++;
+
+  // size must follow = without any white spaces as follows
+  // (1) =300x200
+  // (2) =300x
+  // (3) =x200
+  code = str.charCodeAt(pos);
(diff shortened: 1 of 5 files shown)


puzrin:
@tatsy , it's better to make it as [plugin](https://www.npmjs.org/browse/keyword/markdown-it-plugin), because core should follow [commonmark](http://commonmark.org/) spec. `markdown-it` is specially designed to allow such modifications without upstream change.

tatsy:
@puzrin, thank you for your reply. I understood the ```markdown-it``` design following commonmark. I'll re-implement the function as a plugin. 

What I'm concerned is a code redundancy because the plugin for the size-specified image markup will be almost the same as lib/rules_inline/image.js. Could you give me a piece of advice to prevent the redundancy?

puzrin:
https://github.com/markdown-it/markdown-it/blob/master/lib/index.js#L197-L211

The most critical helpers are already pinned to each instance. Copy the rest, it will not increase total size too much (minified & gzipped)

puzrin:
Or you can just modify renderer - test `title` attr, extract size if exists and add to attributes. That will be much more simple than parser change.

IMHO, renderer patch should not cause conflicts.

puzrin:
Ah, sorry, last advice was wrong, title require quoting. That will work only with this notation: `image`

tatsy:
Oh, I missed the exports of ```helpers``` that you mentioned. It was exactly what I wanted to know. Thank you for your advice.

tatsy:
I'm sorry for rapid-fire questions. Is there any reason why ```lib/helpers/normailze_reference.js``` is not exported in ```lib/helpers/index.js``` ?

puzrin:
No prob. Ask as much as you need. Real devs are always welcome.

That file was not exported, because i was not sure, should it be in helpers or in utils. Forgot to care about it.

Created #42. Copy it temporary, i'll try to do something in next releases.

tatsy:
I checked the created issue. Thank you for your help.

And now, I released a plugin [```markdown-it-imsize```](https://www.npmjs.com/package/markdown-it-imsize) at NPM.

So, please close this PR if everything is fine. Thank you.

puzrin:
you need to refer md from wrapper, instead of require markdown-it directly.

https://github.com/markdown-it/markdown-it-footnote/blob/master/index.js#L43

tatsy:
Thank you for your advice!

I revised my code so that md from wrapper is used in the other functions.

https://github.com/tatsy/markdown-it-imsize/blob/master/index.js#L207-L208

puzrin:
Looks correct.

puzrin:
Moved `normalizeReference` to utils. Will release soon with other minor fixes.

PR #47, opened 2015-02-06
validateLink: expand entities before trimming and lowercasing

opennota:
Logically, the HTML entities should be expanded before the calls to `trim()` and `toLowerCase()`, otherwise there's a possibility that a malicious input can slip past the validation, because the entities can expand to `\n`, `\t`, space and/or capital letters, e.g. `&#0x0a;`, `&#x26;#74;avascript:alert(1)` etc. Now, it doesn't seem to result in a vulnerability (`escapeHTML()` takes care of `&`s), but I wouldn't take such a risk.

### DIFF
--- a/lib/parser_inline.js
+++ b/lib/parser_inline.js
@@ -30,10 +30,10 @@ var _rules = [
 var BAD_PROTOCOLS = [ 'vbscript', 'javascript', 'file' ];
 
 function validateLink(url) {
-  var str = url.trim().toLowerCase();
-
   // Care about digital entities "javascript&#x3A;alert(1)"
-  str = replaceEntities(str);
+  var str = replaceEntities(url);
+
+  str = str.trim().toLowerCase();
 
   if (str.indexOf(':') >= 0 && BAD_PROTOCOLS.indexOf(str.split(':')[0]) >= 0) {
     return false;
--- a/test/fixtures/markdown-it/xss.txt
+++ b/test/fixtures/markdown-it/xss.txt
@@ -29,8 +29,15 @@ Should not allow some protocols in links and images
 
 .
 [xss link](&#34;&#62;&#60;script&#62;alert&#40;&#34;xss&#34;&#41;&#60;/script&#62;)
+
+[xss link](&#74;avascript:alert(1))
+
+[xss link](&#x26;#74;avascript:alert(1))
+
 .
 <p><a href="%22%3E%3Cscript%3Ealert(%22xss%22)%3C/script%3E">xss link</a></p>
+<p>[xss link](Javascript:alert(1))</p>
+<p>[xss link](&amp;#74;avascript:alert(1))</p>
 .
 
 .


puzrin:
"replace entities" changes only html entities, and can't affect `\r\n\t...`. Functions swap will not change anything in this logic.

Also note, we have tests for xss https://github.com/markdown-it/markdown-it/blob/master/test/fixtures/markdown-it/xss.txt

opennota:
> "replace entities" changes only html entities, and can't affect \r\n\t

By `\n` `\t` I meant literal newline and tab stop. `&#x0a;`, `&#x0d;`, `&#09;` will be expanded to literal newline, carriage return and tab stop. `&#74;` will expand to `J`. There's no much point in doing `trim()` and `toLowerCase()` before entity expanding if the expanded URL will contain newlines and capital letters again.

puzrin:
trim() & toLowerCase() are used only to simplify next condition. This validator only check link schema and nothing else. It doesn't care about control characters in the middle of string at all, because that's not needed for security

opennota:
`&#74;avascript:` is a valid schema from the viewpoint of a browser. It's also equal to `javascript:`, but your validator will not catch it.

puzrin:
It will https://github.com/markdown-it/markdown-it/blob/master/test/fixtures/markdown-it/xss.txt#L31

May be i miss something. But could you add test as proof then?

puzrin:
It's possible to get `...href="   javascript:..."` but afaik such links are invalid.

opennota:
Call `validateLink()` directly. It DOES NOT catch `&#74;avascript:`. That that it doesn't result in a XSS is another matter entirely.

puzrin:
Ah, got it, finally, Thanks for you patience :)

Could you add appropriate test to this pr?  It should fail without your fix and ok after it.

opennota:
> It's possible to get ...href=" javascript:..." but afaik such links are invalid.

You're wrong on both counts. Try this:
```
<html>
<body>
  <a href=" javascript:alert(1)">click me</a>
</body>
</html>
```

Also, `encodeURI()` encodes spaces to `%20`, which makes it unusable. But you're right in that it, also, won't be caught by `validateLink()`.

> Could you add appropriate test to this pr? It should fail without your fix and ok after it.

I'll try.

puzrin:
May be name `validateLink` is a bit confusing. It doe's make "browser" link validation, it check, that link willl not be bad after encoding. So, trailing spaces are not problem, but `&#74;avascript:` will be

opennota:
Added tests.

opennota:
> What is this test for ?

Double entity decoding, you know. URLs are passed to `validateLink()` after one round of entity decoding.

puzrin:
I think, it's safe so skip double decoding testing, because it can not happen by design. Or, at least, it should be moved to separate group, because this problem is not `javascript/vbscript:` specific.

Additional normalization tests are here:

- https://github.com/opennota/markdown-it/blob/3a74024d9a3440f27f9521f870b78290c463dd41/test/fixtures/markdown-it/xss.txt#L37
- https://github.com/markdown-it/markdown-it/blob/master/test/fixtures/markdown-it/fatal.txt

opennota:
> I think, it's safe so skip double decoding testing, because it can not happen by design.

It's exactly what happens when you `replaceEntities()` in the `validateLink()` func.

opennota:
Well, I'm a bit tired, and I don't really even use markdown-it, so I think I should stop pestering you. Feel free to close this PR.

puzrin:
I mean, if you doubt, that replaceEntities() can fail and do double decoding, the best place is unit test for this function https://github.com/markdown-it/markdown-it/blob/master/test/utils.js#L30 .

puzrin:
I've merged, because this PR is useful and fixes a real bug. Sorry that it's taken a bit more time than usual.

puzrin:
Released with your fix. https://github.com/markdown-it/markdown-it/commit/0dc415f1d24fbaf65c4bb95e533976a4dc7153a4#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR4

Thanks!

opennota:
> I mean, if you doubt, that replaceEntities() can fail and do double decoding, the best place is unit test for this function https://github.com/markdown-it/markdown-it/blob/master/test/utils.js#L30 .

I believe it belongs to where it is. It's not that I think that `replaceEntities()` can do double decoding, it's that it is actually called *twice*: once in `parseLinkDestination()` (-> `normalizeLink()` -> `replaceEntities()`), and then once inside `validateLink()`.

Issue #86, opened 2015-04-13
Backticked pipes in table cells shouldn't delimit columns


jbt:
I know there was already some stuff in #5 but I don't think it was fully resolved - it looks like 0076a1ebabf92220ec4657536a46ae2da4788733 doesn't actually fix the final case of a pipe char inside backticks that _isn't_ escaped with a backslash

I've tried to do some reading around but I can't find any kind of consensus on what the Right Thing To Do here is:

```markdown
 A |  B  | C 
---|-----|---
 1 | `|` | 3
```

As rendered by GitHub:

 A |  B  | C 
---|-----|---
1 | `|` | 3

However markdown-it interprets the backticked `|` as a cell delimiter and [breaks](https://markdown-it.github.io/#md64=eyJzb3VyY2UiOiIgQSB8ICBCICB8IEMgXG4tLS18LS0tLS18LS0tXG4gMSB8IGB8YCB8IDMiLCJkZWZhdWx0cyI6eyJodG1sIjpmYWxzZSwieGh0bWxPdXQiOmZhbHNlLCJicmVha3MiOmZhbHNlLCJsYW5nUHJlZml4IjoibGFuZ3VhZ2UtIiwibGlua2lmeSI6dHJ1ZSwidHlwb2dyYXBoZXIiOnRydWUsIl9oaWdobGlnaHQiOnRydWUsIl9zdHJpY3QiOmZhbHNlLCJfdmlldyI6Imh0bWwifX0=).

Looks like [marked](https://github.com/chjj/marked) does the same thing, but there's an issue open there: https://github.com/chjj/marked/issues/285 - that hasn't been addressed though so I don't know what the "correct" behaviour should be.

(ref. https://github.com/jbt/markdown-editor/issues/20)

PR #87, opened 2015-04-13
Don't split cells on pipe characters inside backticks.

jbt:
This is for #86. As I said over there I'm not _entirely_ sure whether this should be considered the correct behaviour, but if it is, then this addresses that.

### DIFF
--- a/lib/rules_block/table.js
+++ b/lib/rules_block/table.js
@@ -16,12 +16,17 @@ function escapedSplit(str) {
       max = str.length,
       ch,
       escapes = 0,
-      lastPos = 0;
+      lastPos = 0,
+      backTicked = false,
+      lastBackTick = 0;
 
   ch  = str.charCodeAt(pos);
 
   while (pos < max) {
-    if (ch === 0x7c/* | */ && (escapes % 2 === 0)) {
+    if (ch === 0x60/* ` */ && (escapes % 2 === 0)) {
+      backTicked = !backTicked;
+      lastBackTick = pos;
+    } else if (ch === 0x7c/* | */ && (escapes % 2 === 0) && !backTicked) {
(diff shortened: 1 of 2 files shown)


puzrin:
Looks nice. Could you add one more test - what happens if backtick is opened but not closed?

This patch in theory can do something strange if i place url with backtick into table cell, but i don't think this ever happens in real life.

jbt:
Oh that's a good point actually, I don't quite know what should happen there. GitHub is a bit weird in that case:

```
 A |  B  | C 
---|-----|---
 1 | `| | 3
```

produces...

 A |  B  | C 
---|-----|---
 1 | `| | 3


but

```
 A |  B  | C 
---|-----|---
 1 | ` | | 3
```

(with extra space) produces...

 A |  B  | C 
---|-----|---
 1 | ` | | 3


I'll have a look and see if I can figure out the best way of dealing with that. I don't actually understand why the first case behaves as it does - I'd have thought it should behave like the second case both times.

If it comes to it, does it matter if the first case isn't treated exactly the same as how GitHub does it?

puzrin:
I think, it's not principle to do 100% as gothub does. IMHO, if backtick is not closed, it worth consider it as usual char (as for text without any table)

jbt:
Ok then, I've got another change and test-case that I'll commit and push now

puzrin:
Also, it worth to properly handle escaping for that:

```
\\\`|\\\` => two rows "\`" 
```

jbt:
Ok, I added treatment for the unclosed-backtick case, as well as a test for the escaping example you mentioned

puzrin:
could you squash to single commit?

jbt:
There, all squashed :)

puzrin:
Published. Thanks a lot for PR!

PR #88, opened 2015-04-14
Fixed example for adding target to links

adam-p:
This fixes the sample in `architecture.md` that shows how to add a link target by overriding the default renderer. It was missing a couple returns and had a parameter name typo.


puzrin:
Thanks!

PR #89, opened 2015-04-17
Added smart arrow support to typographer

adam-p:
Here are the additions:

```
--> →
<-- ←
<--> ↔
==> ⇒
<== ⇐
<==> ⇔
```

I [created a plugin](https://github.com/adam-p/markdown-it-typographer-plus) to do this, but then I figured you might like the option of adding it upstream.

### DIFF
--- a/lib/rules_core/replacements.js
+++ b/lib/rules_core/replacements.js
@@ -1,4 +1,4 @@
-// Simple typographyc replacements
+// Simple typographic replacements
 //
 // '' → ‘’
 // "" → “”. Set '«»' for Russian, '„“' for German, empty to disable
@@ -10,6 +10,8 @@
 // ... → … (also ?.... → ?.., !.... → !..)
 // ???????? → ???, !!!!! → !!!, `,,` → `,`
 // -- → &ndash;, --- → &mdash;
+// --> → →; <-- → ←; <--> → ↔
+// ==> → ⇒; <== → ⇐; <==> → ⇔
 //
 'use strict';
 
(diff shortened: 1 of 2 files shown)


puzrin:
1. I would rename your plugin to `markdown-it-smart-arrows`.
2. Take a look at `uc-micro` https://github.com/markdown-it/markdown-it/blob/master/lib/common/utils.js#L249, there are some regexes for better unicode support. `\s` is ok for english only. Those regexes are widely user in [linkifier](https://github.com/markdown-it/linkify-it/blob/master/lib/re.js), but i did not updated typographer yet.
3. `-->` and `<--` will be conflicting with html comments if user enable html. That's a reason why i think it's better to keep arrows in separate plugin. But, personally, i like this rules (i don't use html blocks in my projects).

adam-p:
> 1. I would rename your plugin to markdown-it-smart-arrows.

I thought about that, but I do plan to add more items in the future. On the other hand... I could always create a different plugin for the new stuff. I probably will rename it.

> 2. Take a look at uc-micro

Thanks for pointing me towards that. The i18n-unfriendliness of regexes has always bothered me, and I didn't know about the Unicode libraries (which uc.micro is a subset of). I'll be incorporating that into my code.

> 3. --> and <-- will be conflicting with html comments if user enable html.

Good point. It doesn't matter to me in the context of [Markdown Here](https://github.com/adam-p/markdown-here) since HTML comments get stripped out of email, but... I can see it might matter in the larger sense.

I'll close this PR and stick with the plugin. Thanks for your help.

puzrin:
> I didn't know about the Unicode libraries

Try this ones:

- https://github.com/slevithan/xregexp (no astrals)
- https://github.com/mathiasbynens/regenerate (with astrals, but less features)

Problem is with negative classes, when you need astral characters (surrogates) support. Workaround contains black magick with negative lookaheads, and it's not automated. You can see it at linkifier link i given before.

puzrin:
If you don't object, i'll publish separate arrows plugin with your markup sometime... (if nobody do it before). Similar feature was already suggested in [another PR](https://github.com/markdown-it/markdown-it/pull/25). Seems people need it.

adam-p:
I have no objections. Feel free to use my markup.

PR #106, opened 2015-05-14
Enable configuration of linkify target in config

joshje:
I wanted linkify links to open in a new window, but couldn't see a way of enabling this.

- Passing the config `{ linkify: true }` will work as it does currently
- Passing the config `{ linkify: { target: '_blank' } }` will cause all linkify links to open in a new window

### DIFF
--- a/lib/rules_core/linkify.js
+++ b/lib/rules_core/linkify.js
@@ -18,7 +18,7 @@ function isLinkClose(str) {
 
 module.exports = function linkify(state) {
   var i, j, l, tokens, token, currentToken, nodes, ln, text, pos, lastPos,
-      level, htmlLinkLevel, url, fullUrl, urlText,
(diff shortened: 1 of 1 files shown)


puzrin:
I don't see reasons why this code should be in core.

joshje:
Hi @puzrin. What would be a better way to specify the target for a link generated by linkify? I'm open to suggestions.

puzrin:
You can replace linkifier wrapper plugin, or write one more that will scan link tokens and update targets. 

I doubt, that this setting is very useful in general - many people prefer set target depending on domain (our / foreign), not link source. Also, it's impossible to put in options everything that people can request. General recommendation is to write plugins or do monkeypatching, instead of core change requests .

rlidwka:
I think it's better to override renderer in this case. That's exactly what renderer step is for.

PR #124, opened 2015-06-28
Update README.md

opennota:


puzrin:
IMHO that's not correct because your parser [does not support plugins](https://github.com/opennota/markdown#references--thanks). That's a main reason why markdown-it was written.

I think it will be better to place link in [CM wiki](https://github.com/jgm/CommonMark/wiki/List-of-CommonMark-Implementations) & announce on CM forum.

PR #152, opened 2015-08-24
Added a layer of abstraction so tokens can be directly gotten

funkjunky:
This was so I could write a plugin that could render React components.

Upsides: Cleaner code (map/reduce), any other rendering engine that could use objects could take advantage of the abstraction.

Downside: an additional array creation well use up n memory and cause n allocations where n is the number of tokens generated by the parser.

Unfortunately, I can't seem to run the benchmark, because an autolinker module can't be found.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -57,7 +57,7 @@ function normalizeLink(url) {
     if (!parsed.protocol || RECODE_HOSTNAME_FOR.indexOf(parsed.protocol) >= 0) {
       try {
         parsed.hostname = punycode.toASCII(parsed.hostname);
-      } catch(er) {}
+      } catch(er) { /**/ }
     }
   }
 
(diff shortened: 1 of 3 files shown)


puzrin:
That will never be accepted. Do plugin with custom renderer, instead of such PRs.

funkjunky:
Ok, I understand if you don't feel the benefits out-weigh the cons.
Until I can find the time to transform the changes into a renderer I'll just maintain a forked version.

Btw I'm not sure if you realize this, but the blunt nature of your words comes off as rude. I noticed it in all your comments as well.
I advise against using words such as "never" and "such" in the context you used them. The "never" doesn't add anything, because "this" refers to a pull request that well be different from the next pull request. It well piss off someone who interprets the vague statement as "I well never accept any pull request from you, regardless of justification."

Using the word "such" in the way you used it implies the PR was ill-considered. Which again would infuriate someone with less patience, considering the author of the PR ensured a new test was created for the new code, ensured all tests passed, and that the pedantic code restrictions were met for Travis CI.

PR #157, opened 2015-09-29
Check for the presence of `env` before `env.references`.

blambeau:
Hi!

The following code failed on my computer with `Cannot read property 'references' of undefined`.

```
md = require('markdown-it')('commonmark');
md.parse("Hello you: [link]");
```

I ended up patching markdown-it as shown in the pull request. To be honest, I haven't been able to write a test to reproduce the bug in your test suite, sorry. I wanted your opinion on the exact fix anyway.

### DIFF
--- a/lib/rules_inline/link.js
+++ b/lib/rules_inline/link.js
@@ -95,7 +95,7 @@ module.exports = function link(state, silent) {
     //
     // Link reference
     //
-    if (typeof state.env.references === 'undefined') { return false; }
+    if (typeof state.env === 'undefined' || typeof state.env.references === 'undefined') { return false; }
 
     // [foo]  [bar]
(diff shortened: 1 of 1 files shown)


puzrin:
Second param is mandatory, because can return data. It should not be skipped. Probably, we could throw exception in this case.

blambeau:
Oh, indeed thanks. I should have checked the API more carefully, it's clear that `env` is required there.

kossnocorp:
It would be nice to have an exception. Just ran into the problem and had to use empty block: `md.parse(this.props.value, {})`.

Not sure if I'll have some issues because of that, but at first sight is seems like unnecessary limitation and for me it's even better to assign an empty object as the default value.

PR #158, opened 2015-09-30
adding default_attributes feature to markdown-it renderer

hallvors:
Adding a simple way to define default attributes for given tags in the markdown-it configuration.


puzrin:
Coul you explain when that's needed except `rel="nofollow"`.

puzrin:
@hallvors , i've replied to you in gitter. Let's keep api simple while problem can be solved via plugin.

If you need any help - let me know.

PR #160, opened 2015-10-01
Fix compatibility with closure compiler

### LINKED ISSUE #159: Fix compatibility with closure compiler
- [ ] Update [`mdurl` dependency](https://github.com/markdown-it/markdown-it/blob/master/package.json#L30) to 1.0.1
- [ ] Change the variable name `char` to something different in [`isPunctChar`](https://github.com/markdown-it/markdown-it/blob/master/lib/common/utils.js#L189).

hubgit:
fixes #159


puzrin:
Thanks!

PR #177, opened 2015-11-29
add note about requiring markdown-it

arve0:
I've stumbled upon several plugins ([toc and anchor](https://github.com/MoOx/markdown-it-toc-and-anchor/blob/master/src/index.js#L1), [named headers](https://github.com/leff/markdown-it-named-headers/blob/master/index.js#L4), [anchor](https://github.com/valeriangalliat/markdown-it-anchor/blob/master/index.es6.js#L3)) which require markdown-it directly. In this case, it seems like markdown-it-anchor was the first to require markdown-it, then the others have followed. markdown-it-named-headers doesn't even use `Token`.

Tldr; I think this note is useful for fresh plugin writers.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -18,6 +18,8 @@ Prior to continue, make sure you've read:
     block & inline rules are usually faster.
   - Sometime it's enougth to modify renderer only (for example, to add
     header IDs or `target="_blank"` for the links).
+  - Plugins should use passed arguments to access markdown-it. Plugins should 
+    *never* require markdown-it directly.
 2. Search existing
    [plugins](https://www.npmjs.org/browse/keyword/markdown-it-plugin)
    or [rules](https://github.com/markdown-it/markdown-it/tree/master/lib),


puzrin:
Those plugins require one class from markdown-it. It's not nesessary, but technically possible. I don't think such note will be clear.

puzrin:
I think, you mean, plugins should not use markdown-it as dependency in config. Probably, it's better to write plugin authors, that they should use `token = md.inline.State.push(...)` instead of direct Token require. https://github.com/markdown-it/markdown-it/blob/master/lib/parser_block.js#L122

arve0:
Ok, I can adjust the PR to your suggestions. Is `development.md` the correct place for the note?

puzrin:
https://github.com/markdown-it/markdown-it/commit/b36c22600aee585a2685cac979f224b5e89df900

I've added rephrased note as stub. Feel free to change if needed.

puzrin:
Thanks for your suggestion! May be this note will help to someone.

PR #183, opened 2015-12-31
Add attrReplace

iliakan:
A function to replace attr value for plugins.


puzrin:
IMHO `.attrIndex()` is enougth. Replacing attrs is too specific feature to promote it into API.

PR #186, opened 2016-01-04
Update commonmark spec to 0.24

nikolas:


puzrin:
I guess that should be updated together with good/bad files to be useful. If not 100% of tests pass, it's better to create issue instead of pr

nikolas:
i see, i'm not familiar with the good and bad files, i just use spec.txt. I'll look into it.

puzrin:
@rlidwka please update spec.

nikolas:
0.24 is out now, so I've updated this PR. How are the `good.txt` and `bad.txt` files derived? I don't see them in commonmark.js's test tools: https://github.com/jgm/commonmark.js/tree/master/test

rlidwka:
> How are the good.txt and bad.txt files derived?

They're generated with [specsplit.js](https://github.com/markdown-it/markdown-it/blob/master/support/specsplit.js), so you can run:

```sh
~/markdown-it$ ./support/specsplit.js good /path/to/spec.txt
```

To generate good.txt, and `specsplit.js bad` to generate other one. I think they should be named `passing.txt` and `failing.txt` really, and first one contains spec tests that markdown-it parses correctly and second one contains the rest. Ideally, we'd not have any failing CM tests (i.e. `bad.txt` would be 0 size), but sometimes we do.

Unfortunately, the format of tests was [changed](https://github.com/jgm/CommonMark/commit/2413d436ed98866e9a70fbd4070a5da52dd97cd5) recently, so it has a showball's chance in hell of working properly. So we'll need to update [markdown-it-testgen](https://github.com/markdown-it/markdown-it-testgen) to understand that format.

P.S.: Travis build passes 'cause tests are run on `good.txt` which wasn't updated.

puzrin:
Closing, since change is inconsistent.

rlidwka:
Small errata: turns out, changing `markdown-it-testgen` isn't really necessary.

Updated spec in https://github.com/markdown-it/markdown-it/commit/ab36c448647bea4d5d1c1792206ebb24d4af82ab, thanks for the reminder.

PR #204, opened 2016-02-24
Annotation

SamyPesse:
:construction_worker: WIP

The goal of this PR is to make annotation possible by adding properties `position` and `size` to tokens.

This could be useful for annotating markdowns (my use-case), linting, etc

I wanted to post this PR before working on implementing it for all rules, because I'm not sure that you'll consider merging this, since to work correctly it'll require the rules from plugins to set `position` and `size` of tokens.

### Annotating:

My use-case requires to annotate code (blocks and inline) with `{% raw %}<code>{% endraw %}`, this could be achieved with 2 methods:

- **Method 1**: `Markdown -> [markdown-it.parse] -> tokens -> [annotating transformation] -> tokens -> [markdown-it-mdrenderer] -> Markdown` (using https://github.com/SamyPesse/markdown-it-renderer)
- **Method 2**: `Markdown -> [markdown-it.parse] -> tokens -> [merge raw of tokens using src.slice(token.position, token.position + token.size)] -> Markdown`

**Method 1** is more reliable since it keeps the original source and avoid some `tokens -> markdown` rendering "bugs".

### Progress

- [ ] Blocks
      - [x] Paragraphs
      - [x] Headings
      - [x] LHeadings
      - [x] Tables
      - [x] HR
      - [ ] HTML
      - [x] Code blocks
      - [x] Fences
      - [ ] Blockquote
      - [ ] List
      - [ ] References
- [ ] Inline
      - [x] code / backticks
      - [ ] images
      - [ ] links
      - [ ] emphasis
      - [ ] text
      - [ ] html inline
      - [ ] new line

### DIFF
--- a/lib/rules_block/code.js
+++ b/lib/rules_block/code.js
@@ -4,7 +4,9 @@
 
 
 module.exports = function code(state, startLine, endLine/*, silent*/) {
-  var nextLine, last, token, emptyLines = 0;
+  var nextLine, last, token, emptyLines = 0,
+    pos = state.bMarks[startLine],
+    endPos;
 
   if (state.sCount[startLine] - state.blkIndent < 4) { return false; }
 
@@ -34,11 +36,15 @@ module.exports = function code(state, startLine, endLine/*, silent*/) {
     break;
   }
 
+  endPos = state.bMarks[last] + state.tShift[last];
+
   state.line = last;
 
   token         = state.push('code_block', 'code', 0);
   token.content = state.getLines(startLine, last, 4 + state.blkIndent, true);
   token.map     = [ startLine, state.line ];
+  token.position = pos;
+  token.size = endPos - pos;
 
   return true;
 };
--- a/lib/rules_block/fence.js
+++ b/lib/rules_block/fence.js
@@ -4,7 +4,7 @@
 
 
 module.exports = function fence(state, startLine, endLine, silent) {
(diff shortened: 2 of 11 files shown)


puzrin:
In general:

- Sourcemap feature is really needed, but seems to need serious rewrite of current parser. That's the only reason why we did not implemented it - time limits.
  - We have now lines mapping for block elements https://github.com/markdown-it/markdown-it/blob/6.0.0/support/demo_template/index.js#L164-L179 , that's enougth for specific tasks like sync scroll, but not enougth for more complicated things, related to editing.
- Yes, i'm interested to merge it, if it's done well.

Currently there are 2 big differences with reference implementation:

1. Precalculation of line borders
   - It was fine at the begining, but requires shifting for nested quotes and other blocks. Also can require more memory, when it's difficult to find block end. Still works well until you need to calculate line column offset :)
2. Immutable tokens after pushing to stream. Reference implementation sometime cheat with tail of created ast nodes.
   - Initial idea was to be friendly with JITs and to get speed comparable with C implementation. But we failed :( . Markdown-it is 5 times slower that reference C parser. Probably, we could migrate from tokens to AST without speed loss and simplify things. But that requires a big rewrite to check.

So, it's a big question, what is better: add mapping to existing parser or rewrite it first.

puzrin:
No activity. Let me know if i should reopen it.

PR #220, opened 2016-03-31
remove snake_case from indentFound

slang800:


puzrin:
Sorry, i tend to reject. We use snake_case in many places for internal vars (that's a question of taste, subjective), and i done see reason in partial replace that changes nothing.

slang800:
The idea behind this is to use a naming convention that is consistent within the markdown-it codebase, and with JavaScript as a whole. It does not change any functionality, but does improve readability.

I redid the commit and included it in https://github.com/markdown-it/markdown-it/pull/223

puzrin:
I have opposite opinion - sometime snake_case is more readable for me. We follow node's convention in API, but don't follow it in internal things.

slang800:
Yeah, purely in terms of snake_case vs camelCase, I find snake_case to be easier to read. For example, if I were designing a new language I'd probably follow Python naming semantics and use snake_case for variables and functions.

However, when you start to mix the two it becomes a pain due to the inconsistency. At least personally, I remember variable names as a series of morphemes, rather than characters. So if I read `indent_found` I'll remember it as "indent found" and (in JS) will write it as `indentFound` if I don't remember the specific way it was written first.

puzrin:
I don't see problems with inconsistency. Since this is subjective, let's leave everything as is, without changes.

slang800:
That's interesting that you aren't bothered by the inconsistency. From what I understand, the conversion of familiar words into morphemes is a well-accepted part of the modern [reading model](http://www.pitt.edu/~perfetti/PDF/Coltheart%2005.pdf). So, needing to switch randomly between two ways of writing (like `indent_found` vs `indentFound`) depending upon the context, should place a significant burden on the writer. Do you use tab-completion in your editor or something else that would remove the need to remember how a given identifier is written?

And I don't think that this is subjective at all. The visual representation of code has a direct impact on how easily someone can understand and modify it. See [Brainfuck](https://en.wikipedia.org/wiki/Brainfuck) (or even minified JS) if you need an example of this. We should be able to measure that difference in ease of understanding and pick one way of formatting that is objectively superior.

puzrin:
> Do you use tab-completion in your editor or something else that would remove the need to remember how a given identifier is written?

yes.

Issue #171, opened 2015-11-24
Simple tables vs Complex tables


Martii:
This package doesn't seem to handle complex tables in GFM such as:

 | [![Mozilla Firefox][imgfx]][urlfx] | [![Mozilla SeaMonkey][imgsm]][urlsm] | [![Apple Safari][imgsi]][urlsi] | [![Opera Software Opera][imgoa]][urloa] | [![Microsoft Internet Explorer][imgie]][urlie] | [![Chromium Projects Chromium][imgcm]][urlcm][![Google Chrome][imgce]][urlce] | &emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;
:-----: | :-----: | :-----: | :-----: | :-----: | :-----: | :-----: | :-----: | :-----:
![Linux][imgtux]     | [![Greasemonkey][imggmao]][urlgmao] [![Greasemonkey prc][imggmsf]][urlgmsf]&#10;[![uso - installWith][imgiwi]][urliwi] [![GM_config][imggmc]][urlgmc] | [![Greasemonkey Port][imggpsf]][urlgpsf] [![uso - installWith][imgiwi]][urliwi] [![GM_config][imggmc]][urlgmc] | &ndash; | &ndash; | &ndash; | &ndash;
![Macintosh][imgmac] | [![Greasemonkey][imggmao]][urlgmao] [![Greasemonkey prc][imggmsf]][urlgmsf]&#10;[![uso - installWith][imgiwi]][urliwi] [![GM_config][imggmc]][urlgmc] | [![Greasemonkey Port][imggpsf]][urlgpsf] [![uso - installWith][imgiwi]][urliwi] [![GM_config][imggmc]][urlgmc] | &ndash; | &ndash; | &ndash; | &ndash;
![Windows][imgwin]   | [![Greasemonkey][imggmao]][urlgmao] [![Greasemonkey prc][imggmsf]][urlgmsf]&#10;[![uso - installWith][imgiwi]][urliwi] [![GM_config][imggmc]][urlgmc] | [![Greasemonkey Port][imggpsf]][urlgpsf] [![uso - installWith][imgiwi]][urliwi] [![GM_config][imggmc]][urlgmc]| &ndash; | &ndash; | &ndash; | &ndash;

[...]

PR #223, opened 2016-03-31
fix table rendering

slang800:
I'm working on replacing [marked](https://github.com/chjj/marked) with markdown-it in a formatting tool called [tidy-markdown](https://github.com/slang800/tidy-markdown) (to conform to the commonmark standard).

This PR does 3 things (see commit messages):

- fix inconsistent variable naming, as a follow-up to https://github.com/markdown-it/markdown-it/pull/220#issuecomment-203803325
- fix handling for empty rows at the beginning of the table
- fix truncation of cells from tbody rows that are longer than the thead row

That last thing this PR does is in opposition to what #171 asked for, so I'll elaborate on the reason for opposing #171 and matching the way that marked handles tables...

The way that GitHub.com does this (truncating all rows that are longer than the thead row) is stupid. It hides content that could be important to readers, without so much as a parser error. And that can be a pain when you're not carefully checking what your markdown looks like after compilation. For example, an extra cell jutting out of an otherwise rectangular table, or a missing header is an easy error to spot, while a column missing still leaves the table looking correct and orderly.

Even if you are looking at the output in a graphical/rendered form while you're editing, it's annoying to start a table cell in a new column and not be able to see what you are typing until you add a header for that column.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,4 +1,4 @@
-// Main perser class
+// Main parser class
 
 'use strict';
 
--- a/lib/common/html_re.js
+++ b/lib/common/html_re.js
@@ -2,27 +2,27 @@
 
 'use strict';
 
-var attr_name     = '[a-zA-Z_:][a-zA-Z0-9:._-]*';
+var attrName     = '[a-zA-Z_:][a-zA-Z0-9:._-]*';
 
 var unquoted      = '[^"\'=<>`\\x00-\\x20]+';
-var single_quoted = "'[^']*'";
-var double_quoted = '"[^"]*"';
+var singleQuoted = "'[^']*'";
+var doubleQuoted = '"[^"]*"';
 
-var attr_value  = '(?:' + unquoted + '|' + single_quoted + '|' + double_quoted + ')';
+var attrValue  = '(?:' + unquoted + '|' + singleQuoted + '|' + doubleQuoted + ')';
 
-var attribute   = '(?:\\s+' + attr_name + '(?:\\s*=\\s*' + attr_value + ')?)';
+var attribute   = '(?:\\s+' + attrName + '(?:\\s*=\\s*' + attrValue + ')?)';
 
-var open_tag    = '<[A-Za-z][A-Za-z0-9\\-]*' + attribute + '*\\s*\\/?>';
+var openTag    = '<[A-Za-z][A-Za-z0-9\\-]*' + attribute + '*\\s*\\/?>';
 
-var close_tag   = '<\\/[A-Za-z][A-Za-z0-9\\-]*\\s*>';
+var closeTag   = '<\\/[A-Za-z][A-Za-z0-9\\-]*\\s*>';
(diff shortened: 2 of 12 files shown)


puzrin:
General notes:

1. Please, remove commit with renames.
2. Fix support of missed values can be accepted, but `.trimRight()` is not part of es5 and it worth to replace it with something else for better compatibility with old browsers.
3. "allow tables with tbody rows longer than thead" - i'm not ready to accept this change to mainstream. Until commonmark spec complete about tables, github's style is more preferable for our temporary solution (to not produce "one more standard").
  - Of cause, you can change behaviour via plugin, markdown-it is specially designed to be flexible.
  - I understand that tables can be improved significantly in many ways. But we intentionally keep basic support only to avoid conflicts in future, when tables come to commonmark spec.

slang800:
> i'm not ready to accept this change to mainstream. Until commonmark spec complete about tables, github's style is more preferable for our temporary solution (to not produce "one more standard").

It certainly wouldn't be producing "one more standard" - it's just copying marked rather than GitHub. And given that GitHub's [documentation](https://help.github.com/articles/organizing-information-with-tables/) doesn't even specify what happens when you have an extra cell in a tbody row, I'd be inclined to call it an "unspecified behavior" or (less charitably) a "bug", rather than a standard. It doesn't even seem like redcarpet has [tests](https://github.com/vmg/redcarpet/blob/master/test/markdown_test.rb#L157) for these kinds of tables (unless I'm just not seeing them).

> `.trimRight()` is not part of es5 and it worth to replace it with something else for better compatibility with old browsers.

Sure - added a polyfill.

puzrin:
> it's just copying marked rather than GitHub

IMHO, github can be considered as de-facto standard, `marked` - can not. As i explained before, we did the choice, and there are no plans to change mainstream behaviour about tables until CM spec update. It's better to create a separate plugin for advanced tables.

lib/index.js
```diff
@@ -12,7 +12,12 @@ var ParserInline = require('./parser_inline');
 var LinkifyIt    = require('linkify-it');
 var mdurl        = require('mdurl');
 var punycode     = require('punycode');
+var trimRight    = require('trim-right');
 
+if (!String.prototype.trimRight) {
```
puzrin: I'd like to avoid microdependencies and prototype changes. That's too complex to just check that source string length > 0.

puzrin: PS. we can fix tables bug ourself after weekend.

slang800: > I'd like to avoid microdependencies and prototype changes

Ok - removed the trim part entirely.

> PS. we can fix tables bug ourself after weekend.

The you mean the truncation of cells from tbody rows that are longer than the thead row? I don't mind waiting till Monday if you're busy... switching tidy-markdown over to a new compiler is just a side-project rather than something with a schedule.

puzrin: I mean, that a lot of my comments can look annoying in your eyes :). But that's not intentional and i promis that tables fix will land soon.

puzrin:
Merged tables fix, closing the rest.

PR #224, opened 2016-04-01
fix tables with empty cells at beginning

slang800:
this is a subset of the commits from #223 that aren't controversial

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,4 +1,4 @@
-// Main perser class
+// Main parser class
 
 'use strict';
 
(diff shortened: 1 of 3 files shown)


puzrin:
LGTM.

/cc @rlidwka take a look please.

rlidwka:
@slang800 , this commit seem to be causing trouble for indented tables, see https://github.com/markdown-it/markdown-it/issues/325 .

Also, behavior is different from github, if you copy&paste [test](https://github.com/markdown-it/markdown-it/pull/223/commits/10d6448b220f2d815bc5b814f0cee384179c6099#diff-c6865baeb94f0721338091f8663657a2R477) here in comments for example, last line will fail to render.

So I reverted it in master, not released yet. Please tell me if you have a better idea to solve it, or if there's any code that depends on this feature.

It's still possible to allow empty leading rows in some corner cases, but I'm not sure it's worth breaking compatibility with github in this case.

PR #251, opened 2016-05-30
Add an attrGet method to Token

marijnh:
Since accessing attributes on tokens isn't entirely trivial, it might make sense to make it easy for token-consuming code to get a specific attribute. It certainly would help in ProseMirror, where I'm currently exporting my own utility for this purpose. Since there were already a few different `attr_` methods, I hope one more is okay.

### DIFF
--- a/lib/token.js
+++ b/lib/token.js
@@ -164,6 +164,20 @@ Token.prototype.attrSet = function attrSet(name, value) {
 
 
 /**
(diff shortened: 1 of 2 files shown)


puzrin:
Released 6.0.4, enjoy!

marijnh:
Appreciated! 👍

Issue #258, opened 2016-06-18
Allow for 'scoped' rendering


cbreeden:
This is essentially a feature request, but I feel that it is a far-fetched feature request.  I would like to have an API that would instruct markdown-it to cache the parsing of the previous input.  Then upon a request to render the input with additional modifications, it would be nice that if we told markdown-it where modifications took place it could render only what is necessary in a smart way as opposed to re-rendering the entire document.

Here is an example of something that I have been playing with:

``` javascript
_mdCache = [];
function renderMdChange(src, line) {
    if (_mdCache === [] || !line) {
        _mdCache = md.parse(src, {});
        return md.renderer.render(_mdCache, md.options, {});
    }
    
    const length = _mdCache.length;
    for (let n = 0; n < length; n++) {
        const token = _mdCache[n];
        if (token.map && token.map[0] > line) { break; }    
        if (token.type === "paragraph_open" &&
            token.map && token.map[0] <= line && token.map[1] > line) {
                const newSrc = getLines(src, token.map[0], token.map[1]);
                console.log("Re-parsing: " + _mdCache[n+1].content);
                _mdCache = Array.prototype.concat(
                    _mdCache.slice(0,n),
                    md.parse(newSrc, {}),
                    _mdCache.slice(n+3));
                return md.renderer.render(_mdCache, md.options, {});
        }
    }

    _mdCache = md.parse(src, {});
    return md.renderer.render(_mdCache, md.options, {});    
}
```

What happens here is that I tell markdown-it that there has been a 'small' addition on line number `line`.  Then I scan through the token list seeing if that line change occurs inside a paragraph block.  If that change occurs in a paragraph block, I will just re-render the scope of that paragraph block and place it in our tokenlist cache.  I then send it to the renderer to be processed.

I took the example markdown from <https://markdown-it.github.io/> and added a new paragraph to the end.  When I used this technique and added to the paragraph at the end the rendering time reduced from an average of 6 ms to 1 ms.  I believe that there is more room for improvement as well if you consider using a virtual DOM for an out-put as well, where this will allow rendering to be in place too.

I know there are several issues to consider if attempting to implement something like this:

- Some changes will break blocks, this may not work well with this model
- How to determine large changes like deleting or adding lots of new lines from pasting
- Recalculating 'source maps'
- Rendering can not modify token list (see #259)

I think we could very likely dramatically increase the speed to rendering small changes in markdown as long as we know where these changes occur.  Perhaps this could be an avenue worth considering?

[...]

PR #259, opened 2016-06-18
Add shallow cloning for Tokens to preserve the AST after render

cbreeden:
## About

This is a small patch request to keep to tokens list clean while rendering.  Here is an illustration of the issue.

Source:

``` javascript
var mystr = '``` javascript \n if (foo) { bar; } \n ```';
var tokens = md.parse(mystr, md.options);

console.log(md.renderer.render(tokens, md.options, {}));
console.log(md.renderer.render(tokens, md.options, {}));
```

Before pull request:
```
'<pre><code class="language-javascript"> if (foo) { bar; } \n</code></pre>\n'
'<pre><code class="language-javascript language-javascript"> if (foo) { bar; } \n</code></pre>\n'
```

We see that `renderer.render` will modify the token attributes while processing so if you
feed the token list to `renderer.render` again you will see this attribute twice.

After pull request:
```
'<pre><code class="language-javascript"> if (foo) { bar; } \n</code></pre>\n'
'<pre><code class="language-javascript"> if (foo) { bar; } \n</code></pre>\n'
```

## Application

Markdown-it will become to default markdown rendering engine for vscode 1.3 being released early July it appears.  It allows for a live rendering of markdown as you type.  Instead of re-rendering the entire markdown file, I felt that it would be a worth-while experiment to see if there are cases in which you could only render what has been updated in certain situations.  Please see #258 for more details.

## Benchmark impacts

Before pull request:
```
Sample: block-fences.md (72 bytes)
 > commonmark-reference x 80,532 ops/sec ±0.24% (86 runs sampled)
 > current x 172,560 ops/sec ±1.13% (86 runs sampled)
 > current-commonmark x 133,014 ops/sec ±0.75% (91 runs sampled)
 > markdown-it-2.2.1-commonmark x 182,824 ops/sec ±1.23% (85 runs sampled)
 > marked-0.3.2 x 309,737 ops/sec ±0.94% (88 runs sampled)
```

After pull request:
```
Sample: block-fences.md (72 bytes)
 > commonmark-reference x 77,697 ops/sec ±0.26% (90 runs sampled)
 > current x 173,375 ops/sec ±1.05% (82 runs sampled)
 > current-commonmark x 180,919 ops/sec ±1.28% (91 runs sampled)
 > markdown-it-2.2.1-commonmark x 174,600 ops/sec ±0.95% (88 runs sampled)
 > marked-0.3.2 x 296,704 ops/sec ±0.92% (82 runs sampled)
```

It appears to have minimal impact on fenced code blocks.

### DIFF
--- a/lib/renderer.js
+++ b/lib/renderer.js
@@ -29,7 +29,10 @@ default_rules.code_block = function (tokens, idx /*, options, env */) {
 
 
 default_rules.fence = function (tokens, idx, options, env, slf) {
-  var token = tokens[idx],
+  // We will make modifications of the attributes object of this
+  // token while processing, so we obtain a shallow clone as to
+  // prevent modifying the token in the token list.
+  var token = tokens[idx].clone(),
       info = token.info ? unescapeAll(token.info).trim() : '',
       langName = '',
       highlighted;
--- a/lib/token.js
+++ b/lib/token.js
@@ -194,4 +194,28 @@ Token.prototype.attrJoin = function attrJoin(name, value) {
 };
 
 
+/**
+ * Token.clone()
+ *
+ * Obtain a shallow clone of the token.  You can use this while rendering to
+ * prevent modifying the token list while rendering.
+ */
+
+Token.prototype.clone = function clone() {
+  var token = new Token(this.type, this.tag, this.nesting);
+
+  token.attrs = this.attrs;
+  token.level = this.level;
+  token.children = this.children;
+  token.content = this.content;
+  token.map = this.map;
+  token.markup = this.markup;
+  token.info = this.info;
+  token.meta = this.meta;
+  token.block = this.block;
+  token.hidden = this.hidden;
+
+  return token;
+};
+
 module.exports = Token;


cbreeden:
It also should be noted that this is a breaking change, so it probably shouldn't be merged without considerable thought.

puzrin:
Tokens stream must be immutable on render phase. Could you point me a line in source that breaks this principle?

cbreeden:
It happens [in this call](https://github.com/markdown-it/markdown-it/blob/master/lib/renderer.js#L39).  When we [copy the token](https://github.com/markdown-it/markdown-it/blob/master/lib/renderer.js#L32) we essentially have only taken a pointer since objects are mutable, and then then when we modify the `attrs`, being a list, is also mutable and so all of these changes follow through to the token stream.  At least this is my understanding.  I verified this using a debugger.

puzrin:
Got it. That's a bug that should be fixed. But i'm not sure it's needed to add `.clone()`. Created a separate issue to think about. Looks like `attrJoin()` is a design mistake.

puzrin:
Found. `.attrJoin()` is ok, but is was a mistake to use it in renderer. Need to rewrite that function without `.attrJoin` and that will work without breaking changes.

cbreeden:
I agree, using a `.clone()` is probably overkill.  I'll go ahead and close this, but I won't delete the branch if for now if we need to reference it for some reason later on.

PR #276, opened 2016-08-12
fence renderer: fix concat of class array

arve0:
Before:

    <pre c="l"><code>...

Now:

    <pre class="someClass language-python><code>...

Bug introduced in b7c868b64b710c8e706191d89a6dd80fcbbc9166


rlidwka:
Merged both PRs, thanks!

PR #293, opened 2016-10-11
Add `start` attribute to ordered list items, on the first one

barm:
I ran into an issue today where I was trying to apply custom styling to an ordered list that was split over multiple separately rendered markdown nodes. I decided to use a pseudo element with `content: attr(start)`, but this attribute didn't exist on `ol`s where the value would be `1` (because it's the default). 

If we remove the guard against adding `start=1`, we make it easier to implement custom list styles and there should be no functional change (so far as I am aware). Any reason why we should not explicitly set the start to `1`?

### DIFF
--- a/lib/rules_block/list.js
+++ b/lib/rules_block/list.js
@@ -176,9 +176,7 @@ module.exports = function list(state, startLine, endLine, silent) {
 
   if (isOrdered) {
     token       = state.push('ordered_list_open', 'ol', 1);
-    if (markerValue !== 1) {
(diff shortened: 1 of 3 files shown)


puzrin:
> Any reason why we should not explicitly set the start to `1`?

Compatibility with commonmark reference renderer output ?

puzrin:
You could add "plugin" to scan all `ordered_list_open` tokens, and add start to next if not exists. I'd like to avoid data modification for specific user requests, if possible.

barm:
Yeah that's totally reasonable, thanks!

PR #302, opened 2016-11-27
fallback to reference if a link is not valid

ujifgc:
This commit makes martdown-it pass a new example introduced in https://github.com/jgm/CommonMark/commit/cfc84164475d3bec8be9482c21a705adc93a54f5

```
[foo](not a link)

[foo]: /url1
.
<p><a href="/url1">foo</a>(not a link)</p>
```

Ref: https://github.com/jgm/CommonMark/issues/427

### DIFF
--- a/lib/rules_inline/link.js
+++ b/lib/rules_inline/link.js
@@ -20,7 +20,8 @@ module.exports = function link(state, silent) {
       href = '',
       oldPos = state.pos,
       max = state.posMax,
-      start = state.pos;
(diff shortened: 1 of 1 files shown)


puzrin:
Thanks, but our approach is to update spec version with all changes & tests, instead of separate modifications. See git log for details.

I'll keep this open for tracking.

rlidwka:
Merged, spec updated to 0.27. Thanks!

PR #312, opened 2016-12-13
Add Output Option to CLI

mjbvz:
Adds an option to specify the output file when using markdown-it from the command line.

I'm on the VScode team and in our docs, we currently recommend that users install marked for converting markdown to html: https://code.visualstudio.com/Docs/languages/markdown#_compiling-markdown-into-html . I'd like to use markdown-it in this example instead, but correctly setting up a task to use markdown-it this is not as easy as it is with marked. Having a command line output file option fixes this

### DIFF
--- a/bin/markdown-it.js
+++ b/bin/markdown-it.js
@@ -42,6 +42,11 @@ cli.addArgument([ 'file' ], {
   defaultValue: '-'
 });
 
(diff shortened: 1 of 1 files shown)


puzrin:
Ok. Looks reasonable.

puzrin:
Published

PR #343, opened 2017-03-14
add LF to an empty blockquote rendering (cosmetic)

ujifgc:
This way it's more consistent with the reference rendering and does not require a special treatment of spec.txt for testing.

### DIFF
--- a/lib/renderer.js
+++ b/lib/renderer.js
@@ -236,6 +236,9 @@ Renderer.prototype.renderToken = function renderToken(tokens, idx, options) {
           //
           needLf = false;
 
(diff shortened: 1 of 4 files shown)


puzrin:
Spec can not define "reference html formatting", and IMHO this LF should not be there. I had more disagreements about formatting in the past, but most of those were fixed. This one still stay there. I think it's better to leave things as is.

puzrin:
Close - rejected:

- i don't like how spec generates `\n` on blockquotes. That's not consistent with other blocks.

We can discuss this again if new reasons appears. Feel free to kick me with any questions. Sorry for rejecting, i respect your time and your intent to make markdown-it better.

Issue #351, opened 2017-04-29
Why are references required?


martinheidegger:
I just used `.parse` like this:

```javascript
const md = new MarkdownIt()
const tokens = md.parse('# Hi!\n\n Fully test [hello], [hosy](http://google.com) ')
```

and it showed this nice error:

```
not ok Cannot read property 'references' of undefined
    stack: |
      Array.link (node_modules/markdown-it/lib/rules_inline/link.js:101:25)
      ParserInline.tokenize (node_modules/markdown-it/lib/parser_inline.js:135:22)
      ParserInline.parse (node_modules/markdown-it/lib/parser_inline.js:163:8)
      Array.inline (node_modules/markdown-it/lib/rules_core/inline.js:10:23)
      Core.process (node_modules/markdown-it/lib/parser_core.js:51:13)
      MarkdownIt.parse (node_modules/markdown-it/lib/index.js:523:13)
      index.js:43:23
      module.exports (index.js:41:10)
      Test.<anonymous> (test/index.js:5:10)
    at:
      line: 101
      column: 25
      file: node_modules/markdown-it/lib/rules_inline/link.js
      function: Array.link
    type: TypeError
    test: simple
    source: |
      if (typeof state.env.references === 'undefined') { return false; }
```

which kindly disappears if you pass in `references: {}` 

```javascript
const md = new MarkdownIt()
const tokens = md.parse('# Hi!\n\n Fully test [hello], [hosy](http://google.com) ', {references: {}})
```

Why is this necessary? Couldn't an error be shown more clearly? Should there be a document that explains that? _Couldn't find one ... :blush:_

puzrin:
https://markdown-it.github.io/markdown-it/#MarkdownIt.parse

See API docs. IMHO it has enougth details. Could you explain the sequence, how you found `.parse()` method and decided that second param is optional?

martinheidegger:
@puzrin I have used many parsers (and markdown-it) in the past and remembered the `parse` function but simply didn't remember that `env` is required _(my mistake, totally! :sweat_smile: )_. But it does work with env being `null` or `undefined` until you run into a case like the one mentioned above where it tries to lookup a link.

puzrin:
In 99% users need `.render()`. If someone decide to use low level api, he is expected to understend well what he does (read docs, surf sources and so on). Of cause, mistakes are possible but i think, there are no needs now to add more checks into `.parse()`.

Can we close this or do you have objections?

[...]

PR #353, opened 2017-05-05
better check for property 'references' in js object

kospl:
fix for javascript error 'Cannot read property "references" from
undefined'

### DIFF
--- a/lib/rules_block/reference.js
+++ b/lib/rules_block/reference.js
@@ -183,7 +183,9 @@ module.exports = function reference(state, startLine, _endLine, silent) {
   // Reference can not terminate anything. This check is for safety only.
   /*istanbul ignore if*/
   if (silent) { return true; }
-
+  if (typeof state.env === 'undefined') {
+    state.env = {};
+  }
   if (typeof state.env.references === 'undefined') {
     state.env.references = {};
   }
--- a/lib/rules_inline/image.js
+++ b/lib/rules_inline/image.js
@@ -95,7 +95,7 @@ module.exports = function image(state, silent) {
     //
     // Link reference
     //
-    if (typeof state.env.references === 'undefined') { return false; }
+    if (typeof state.env === 'undefined' || typeof state.env.references === 'undefined') { return false; }
 
     if (pos < max && state.src.charCodeAt(pos) === 0x5B/* [ */) {
       start = pos + 1;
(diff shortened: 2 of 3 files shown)


puzrin:
I don't like this PR. It adds tons of checks for param that is mandatory.

kospl:
Maybe I'm wrong, but I don't see any sign that env param is mandatory.

For example Google skips it successfully in their example: https://github.com/googlesamples/md2googleslides/blob/master/src/extract_slides.js#L80
which leads to `Cannot read property "references" from undefined` in some cases. 
I just added this check for those cases.

Maybe there's better solution for this problem, I cannot disagree.

puzrin:
I'm not author of that code and have no ideas why it's done such way. Probably that's a bug.

See docs https://markdown-it.github.io/markdown-it/#MarkdownIt.parse. It never says that param is optional and explains what should be passed.

puzrin:
Also see #351

kospl:
I see docs say it is mandatory, but it never checked in code for being undefined. That should be fixed. 
Is it ok to add `env = env || {};` to `MarkdownIt.prototype.parse` like it's done in `MarkdownIt.prototype.render`?

puzrin:
No. This parameter is optional in `.render()` but mandatory in `.parse()`.

kospl:
Then maybe 
`
  if (typeof env === 'undefined') {
    throw new Error('env parameter is mandatory');
  }
`
?

puzrin:
You try to discuss how to change things prior to decide iif those changes are needed or not.

I think that's not critical and i don't like to change code just for fun. See #351.

digitalmoksha:
I have to agree with @kospl on this one:  is the guard required? no.  Is it good practice, defensive programming?  yes.  particularly for a public entry point.  And while it may be mentioned in the docs, documentation can usually be dense to read.  So I think the guard is useful.

puzrin:
@digitalmoksha but `.parse()` is a low level API, not expected to be called by regular users. It's even hidden in readme. And `.render()` has some additional checks.

digitalmoksha:
@puzrin You're right, `parse` is marked internal, and so my previous comment doesn't really apply.

puzrin:
Close as dupe of #351 &  to avoid open issues/pr with "endless" time.

It's already labeled to be search-able sometime later. If new info about "how this bug happen" appears, feel free to update.

This issue can be revived anytime when more arguments appear.

PR #391, opened 2017-08-04
Benchmark

arve0:
Fix `./benchmark/benchmark.js`. Currently errors with:

```
module.js:471
    throw err;
    ^

Error: Cannot find module '../../extra/lib/node_modules/commonmark'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/arve/git/markdown-it/benchmark/implementations/commonmark-reference/index.js:3:18)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
```

Also ignores `.DS_Store` files, which Mac OS X adds to directories when browsing with Finder. You may ignore 6b6c084 if you do not care about Mac dev users.

### DIFF
--- a/benchmark/benchmark.js
+++ b/benchmark/benchmark.js
@@ -16,6 +16,7 @@ var IMPLS = [];
 
 
 fs.readdirSync(IMPLS_DIRECTORY).sort().forEach(function (name) {
+  if (name === '.DS_Store') { return; }
   var file = path.join(IMPLS_DIRECTORY, name);
   var code = require(file);
 
--- a/benchmark/implementations/markdown-it-2.2.1-commonmark/index.js
+++ b/benchmark/implementations/markdown-it-2.2.1-commonmark/index.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var md = require('../../extra/lib/node_modules/markdown-it')('commonmark');
+var md = require('markdown-it')('commonmark');
 
 exports.run = function (data) {
   return md.render(data);
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ coverage/
 demo/
 apidoc/
 *.log
+.DS_Store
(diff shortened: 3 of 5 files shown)


arve0:
Also `markdown-it-2.2.1-commonmark` is really just current, I believe.

benchmark/benchmark.js
```diff
@@ -16,6 +16,7 @@ var IMPLS = [];
 
 
 fs.readdirSync(IMPLS_DIRECTORY).sort().forEach(function (name) {
+  if (name === '.DS_Store') { return; }
```
puzrin: I could understand such file in .gitignore (don't know what it means), but placing such condition into code looks wrong.

arve0: `fs.readdirSync` finds the file, and tries to require it on line 21.

Might want to read only directories instead.

puzrin:
I don't remember details, but it seems you just need to run `make benchmark-deps` instead.

https://github.com/markdown-it/markdown-it/blob/master/Makefile#L97

arve0:
> run make benchmark-deps instead

Ahh, thanks! I see that it's noted in the README.md, but I read CONTRIBUTING.md.

benchmark/implementations/markdown-it-2.2.1-commonmark/index.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-var md = require('../../extra/lib/node_modules/markdown-it')('commonmark');
+var md = require('markdown-it')('commonmark');
```
puzrin: As far as i see, this benchmark is really for 2.2.1 version, so this change is invalid.

I'd suggest to close PR, because `make benchmark-deps` should help.

PR #470, opened 2018-06-25
add punycode to dependencies

akonwi:
`punycode` is required in `/lib/index.js` but it's not declared in the package.json.

This is an issue for me because I'm using rollup to bundle my library and it can't find `punycode` since it's not a declared dependency and isn't in my `node_modules/`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "entities": "~1.1.1",
     "linkify-it": "^2.0.0",
     "mdurl": "^1.0.1",
+    "punycode": "^2.1.1",
     "uc.micro": "^1.0.5"
   },
   "devDependencies": {


ryan-copperleaf:
https://github.com/calvinmetcalf/rollup-plugin-node-builtins?

coolme200:
https://github.com/markdown-it/markdown-it/pull/513   https://github.com/bestiejs/punycode.js/issues/86

should not depd 2.x

puzrin:
As far as i understand, rollup has appropriate plugin, to process node builtins right. Any other reasons to not close this PR?

akonwi:
Yea this is moot now. I should have closed it a while ago.

Thanks

fiznool:
It’s worth noting that punycode has been slated as deprecated for some time
now in node. The official node docs suggest to use punycode.js instead.

PR #476, opened 2018-07-17
Make --no-html have an effect in CLI

ryan-copperleaf:
HTML is a surprising default, too.


puzrin:
Thanks!

About enabled html. Since CLI is not supported (in sync with library), it has all features enabled bu default, and then some options ware added. Consider CLI as "just for fun" feature and use with cautions.

PR #485, opened 2018-08-14
fix: check if state.env is undefined

sublimino:
This was throwing an error when used from https://github.com/gsuitedevs/md2googleslides

### DIFF
--- a/lib/rules_inline/link.js
+++ b/lib/rules_inline/link.js
@@ -98,7 +98,7 @@ module.exports = function link(state, silent) {
     //
     // Link reference
     //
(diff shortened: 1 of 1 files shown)


puzrin:
`state.env` can NOT be undefined. If you fall into such situation, that means error happened on earlier stage. Check in this place will add more confusion (later) than fixes.

tageecc:
@puzrin hey,im got this error when i was used plugin markdown-it-footnote
image

this is my code:
```
import MarkdownIt from 'markdown-it';
import footnote from 'markdown-it-footnote';
const Markdown = new MarkdownIt();
Markdown.use(footnote);
export default Markdown;
```

PR #488, opened 2018-08-31
Option to turn on smartquotes without typographer

m93a:
I think that smartquotes are cool but I don't like the other replacements. Wouldn't it be nice to have a toggle just for smartquotes?


puzrin:
IMHO that's subjective and not enougth for breaking API changes.

Just use `.disable('replacements')`.

m93a:
That's not an API-breaking change – markdown-it would perform identically unless you set options to `{typographer: false, smartQuotes: true}`, then it would turn on smartquotes without typographic replacements. However thanks for the tip about `.disable('replacements')`, I didn't know that was possible!

PR #493, opened 2018-09-18
Add support for modern node versions

amilajack:


puzrin:
IMHO that does not add any value.

PR #498, opened 2018-10-08
Add event API with events willparse and parsed

m93a:
This PR adds API for events. It works the same way it does in DOM (but it doesn't support bubbling/catching). For now, only two events are supported: `willparse` and `parsed`, both events are dispatched when the function `md.parse` (or `md.parseInline`) is called before and after the parsing respectively.

The events are really useful when making an auto-numbering extension, so that you can reset the number before each rendering, or generally for every extension that needs to manage some document-wide data that should be reset with each rendering. Also, they make the `markdown-it` library more transparent to the external code.

New methods:
```typescript
function addEventListener(name: string, fn: (e: Event) => void);
function removeEventListener(name: string, fn: (e: Event) => void);
function dispatchEvent(name: string, props: object);
```

Example usage:
```javascript
md = markdownit();
md.addEventListener('parsed', (e) => updateMyStuff());
md.render('*a*'); // event called
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -13,6 +13,9 @@ var LinkifyIt    = require('linkify-it');
 var mdurl        = require('mdurl');
 var punycode     = require('punycode');
 
+/* global Symbol */
+var EVENTS = typeof Symbol === 'function' ? Symbol('events') : '__events';
+
 
 var config = {
   'default': require('./presets/default'),
@@ -341,6 +344,8 @@ function MarkdownIt(presetName, options) {
   this.helpers = utils.assign({}, helpers);
 
 
+  this[EVENTS] = {};
+
   this.options = {};
   this.configure(presetName);
 
@@ -520,7 +525,11 @@ MarkdownIt.prototype.parse = function (src, env) {
 
   var state = new this.core.State(src, this, env);
 
+  this.dispatchEvent('willparse', { state: state });
+
   this.core.process(state);
+
+  this.dispatchEvent('parsed', { state: state });
 
   return state.tokens;
 };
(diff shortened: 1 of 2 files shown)


m93a:
I don't believe it's possible to test that one line Coveralls marks as missing coverage…

puzrin:
Sorry, events API does not fit into existing architecture (and can trash performance). You can get similar effect by adding new fake rules or by replaying tokens with your own code.

m93a:
I don't really understand why events should be any slower than fake rules:
 * in case of a fake rule, you push a callback to an array, and with each subsequential parsing a State gets constructed, the array of rules is iterated and each rule is called.
 * in case of an event, you push a callback to an array, and with each subsequential parsing an Event gets constructed, the array of events is iterate and each callback is called.

The difference is that with fake rules you have to use a block rule, correctly place it in the rule list (and hope it doesn't get overriden by any other extension), and shortcircuit it by `state.pos = state.posMax`. This is more prone to errors and requires deeper understanding of the whole library. And the point of libraries is not having to understand how exactly does it work.

So could you explain to me, please, why in particular did you find this PR not worth merging?

puzrin:
I'm very conservative about API changes, because it can be cluttered very quickly, if we start add  everything anyone can imagine. For significant changes, 2 conditions should be satisfied:

1. Desired result can NOT be reached in any other way.
2. This change is needed for many people.

Your PR does not seems to satisfy any (or may be i could not understand).

Also, as a side note, it looks strange to have 2 different approaches for extensions. But the main reasons of decline are above. Existing design is stable and self-sufficient, with all known limitations. If someone decide to bring a new breath into markdown parsers, all info is collected in https://github.com/markdown-it/markdown-ast-spec.

m93a:
Thanks for your reply. If you're really meticulous about the first rule, then there's nothing I can do to make my PR pass it. You definitely can create fake rules (or even monkey-patch the `parse` method) but it feels wrong and overly complicated to me.
However, as the PR doesn't comply with rule 1, I can always make an extension for this, so hey, let's not merge this.

puzrin:
I'd suggest you to start using parser, because first impression may be wrong and writing complicated plugins is not trivial thing. If after several months, when you understand well all details, you still think that changes are really needed, we can return to your proposal again. To save your time - please create issue first, to discuss details and make sure if PR will be merged.

PR #501, opened 2018-10-10
WIP: Escape special chars

SISheogorath:
Right now, the copyright sign as well as other special char 
specifications like `(tm)`, `(r)` and `(p)` are not escapable.

This patch adds the ability to escape them, simply by checking for a `\` 
(Backslash) in front of the first bracket (`\(tm)` or `\(c)`).

Maybe not the best way, but works :crossed_fingers:


puzrin:
I'm not sure this works because escaping should gone before (in current design/implementation). Could you add tests?

puzrin:
Also, as far as i see, existing tests fail. That should be fixed too.

SISheogorath:
Made the existing tests working locally, just to realize that escaping isn't working ^^

PR #509, opened 2018-10-27
Fix language detection with filename

g0tt:
Enables to detect language of code blocks such as:

\`\`\`[lang]:[filename]
foo 
bar
\`\`\`


puzrin:
I'm not sure it's good idea to add such customizations to upstream. Could you explain your reasons?

puzrin:
Close - out of project scope. Write plugin instead.

PR #513, opened 2018-11-20
fix: pkg miss depd punycode

coolme200:
https://github.com/bestiejs/punycode.js/issues/86

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
     "entities": "~1.1.1",
     "linkify-it": "^2.0.0",
     "mdurl": "^1.0.1",
-    "uc.micro": "^1.0.5"
+    "uc.micro": "^1.0.5",
+    "punycode": "^1.4.1"
   },
   "devDependencies": {
     "ansi": "^0.3.0",


puzrin:
Can we close this since original PR not actual and been closed too? Node is about to deprecate embedded punicode sometime, but it still works, and i don't like to change working things.

coolme200:
we depd in react component,  uglify-js compress will failed because of punicode is code with es6

puzrin:
IMHO you have the same problem as author of previous PR - should tune your bundler, to properly stub builtin packages.

7weapon:
I also encountered the same problem, probably when will this pr be merged

laurent22:
For those who come across this post, in my case I solved it for React Native using this: `npm install --save-dev rollup-plugin-node-builtins`

todoubaba:
Same problem when `import * as markdownit from 'markdown-it';` to angular, see https://github.com/angular/angular-cli/issues/9508#issuecomment-391490414

philbates35:
I just hit this as well on Node 14, kinda confusing to explain...

I had a webpack plugin that has punycode v2 as an explicit dependency so v2 was the version in `node_modules/punycode/`. If it wasn't for that webpack plugin, no doubt I'd have hit the "module punycode not found" error - but instead because there is no explicit dependency on `punycode` package in `markdown-it`, it was pulling in the v2 found at `node_modules/punycode/`.

This caused my bundle to blow up in IE11 because punycode v1.4.1 is the last version that works in IE11.

I fixed IE11 support by explicitly adding `"punycode": "^1.4.1",` to my `package.json`, because that forced `node_modules/punycode/` to be v1.4.1 which is the version now being pulled into markdown-it.

If this MR was merged and the dependency made explicit, it wouldn't have had to spend the last few hours working our what was happening, so this MR gets a :+1: from me.

dpeter99:
Hi!
I jsut got an issue related to this while trying to use the package on Deno. They have a compatibility layer for node but they do not polyfill any deprecated APIs. This does mean that on deno it is immpossible to run this package, as punycode is missing.

It would be really awesome if this could be fixed!

JasonGore:
Also running into this as an issue. I just had punycode randomly resolve to a 2.x version that was coincidentally installed by something else. This should be added as a declaration to the `package.json`, especially now that `punycode` in Node has long been deprecated and can no longer be considered "builtin".

barbalex:
Running into same problem:
```

 ERROR #98124  WEBPACK

Generating development JavaScript bundle failed

Can't resolve 'punycode' in 'C:\Users\alexa\apf2\.yarn\cache\markdown-it-npm-13.0.1-dcb1a8675c-faf5891d38.zip\node_modules\markdown-it\lib'

If you're trying to use a package make sure that 'punycode' is installed. If you're trying to use a local file make sure that the path is correct.

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "punycode": require.resolve("punycode/") }'
        - install 'punycode'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "punycode": false }

File: .yarn\cache\markdown-it-npm-13.0.1-dcb1a8675c-faf5891d38.zip\node_modules\markdown-it\lib\index.js:14:19

failed Building development bundle - 72.482s
ERROR in ./.yarn/cache/markdown-it-npm-13.0.1-dcb1a8675c-faf5891d38.zip/node_modules/markdown-it/lib/index.js 14:19-38
Module not found: Error: Can't resolve 'punycode' in 'C:\Users\alexa\apf2\.yarn\cache\markdown-it-npm-13.0.1-dcb1a8675c-faf5891d38.zip\node_modules\markdown-it\lib'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "punycode": require.resolve("punycode/") }'
        - install 'punycode'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "punycode": false }
 @ ./.yarn/cache/markdown-it-npm-13.0.1-dcb1a8675c-faf5891d38.zip/node_modules/markdown-it/index.js 4:0-34
 @ ./src/components/Projekte/Daten/CurrentIssue/index.js 14:0-37 22:19-29
 @ ./src/components/Projekte/Daten/index.js 24:0-42 183:48-60
 @ ./src/components/Projekte/ProjektContainer.js 17:0-28 122:40-45 273:44-49
 @ ./src/components/Projekte/index.js 27:0-50 202:40-56 234:38-54 247:39-55
 @ ./src/pages/Daten.js 15:0-46 131:62-70
 @ ./.cache/_this_is_virtual_fs_path_/$virtual/async-requires.js 16:11-18:5
 @ ./.cache/app.js 17:0-52 30:87-34:1 33:29-42 36:27-40 30:0-34:2
```

Please solve this!

barbalex:
It makes no sense to _not_ declare dependencies. Only creates problems. These problems can be hard to solve, depending on the dev/build environment.

andymac4182:
I understand that there is config you can do for some bundlers to resolve this. If we could move to using the package that would help have less foot guns for the library and make it easier to use. 

Please move to using the npm library instead of the other polyfills. At the moment I am hand patching the lib to handle this on my side.

PR #570, opened 2019-07-09
ADD: support for plugins on the CLI

izodmungo:
Just wanted to add in support for plugins from the CLI.


puzrin:
Existing CLI is for very limited use only (mostly test/demo). There were no plan to support plugins because it's not clear how to pass all possible options in unified format. Do you have solution for that?

puzrin:
Closing. Do advanced CLI in separate package please. If it become stable and useful for many people, we can consider merge it to upstream.

PR #575, opened 2019-07-23
 refactor:autolink rule

LinFeng1997:


puzrin:
Please, don't PR here incomplete things and without explanation why change needed.

LinFeng1997:
@puzrin I want to make this piece of code more readable,thank you

puzrin:
1. Your code has inlined functions, that will cause performance loss.
2. Existing code is good enough, no changes needed. Please avoid changes "just for fun".

LinFeng1997:
@puzrin 
1. I agree with you that inline functions do cause a little loss of performance.
2. I think the code for AUTO_LINK and EMAIL_LINK is duplicated, this should be extracted as a function

puzrin:
Sorry, i don't agree. This code is good as is.

LinFeng1997:
@puzrin Well, you're the master of this code, but I'm still insisting that the logic for dealing with AUTO-LINK and EMAIL-LINK should be extracted as a function.

PR #580, opened 2019-07-31
Linkify inline

### LINKED ISSUE #38: linkify doesn't accept urls with underscores
[Example](https://markdown-it.github.io/#md64=eyJzb3VyY2UiOiJodHRwOi8vbWVkaWEuY2NjLmRlL2Jyb3dzZS9jb25ncmVzcy8yMDE0LzMxYzNfLV82MzczXy1fZW5fLV9zYWFsXzZfLV8yMDE0MTIyOTE2MDBfLV90aGVfb25seV90aGluZ193ZV9rbm93X2Fib3V0X2N5YmVyc3BhY2VfaXNfdGhhdF9pdHNfNjQweDQ4MF8tX29saWFfbGlhbGluYS5odG1sIiwiZGVmYXVsdHMiOnsiaHRtbCI6ZmFsc2UsInhodG1sT3V0IjpmYWxzZSwiYnJlYWtzIjpmYWxzZSwibGFuZ1ByZWZpeCI6Imxhbmd1YWdlLSIsImxpbmtpZnkiOnRydWUsInR5cG9ncmFwaGVyIjp0cnVlLCJfaGlnaGxpZ2h0Ijp0cnVlLCJfc3RyaWN0IjpmYWxzZSwiX3ZpZXciOiJodG1sIn19)
One has to escape the underscores in the url but I don't think that this is the expected behavior.

### LINKED ISSUE #518: Linkify terminating url - requires upgrade?
Hi,

When I have markdown:
`https://www.riotinto.com/documents/_Iron%20Ore/Port_Handbook_Final_-_July_2016(2).pdf`
markdown-it seems to terminate the link at `https://www.riotinto.com/documents/_`
See: [example](https://markdown-it.github.io/#md3=%7B%22source%22%3A%22%5Bclick%20me%5D%28https%3A%2F%2Fwww.riotinto.com%2Fdocuments%2F_Iron%2520Ore%2FPort_Handbook_Final_-_July_2016%282%29.pdf%29%5Cn%5Cnhttps%3A%2F%2Fwww.riotinto.com%2Fdocuments%2F_Iron%2520Ore%2FPort_Handbook_Final_-_July_2016%282%29.pdf%22%2C%22defaults%22%3A%7B%22html%22%3Afalse%2C%22xhtmlOut%22%3Afalse%2C%22breaks%22%3Afalse%2C%22langPrefix%22%3A%22language-%22%2C%22linkify%22%3Atrue%2C%22typographer%22%3Atrue%2C%22_highlight%22%3Atrue%2C%22_strict%22%3Afalse%2C%22_view%22%3A%22html%22%7D%7D)

image

I've tried this in the the linkify-it playground and it [seems to work](http://markdown-it.github.io/linkify-it/#t1=https%3A%2F%2Fwww.riotinto.com%2Fdocuments%2F_Iron%20Ore%2FPort_Handbook_Final_-_July_2016(2).pdf) making me think that the latest release 2.1.0 of linkify-it has fixed the issue I'm seeing.

Would it be possible to upgrade and rebuild the markdown-it dependency?

edemaine:
Fix #518 and fix #38 by running linkify during the inline phase, so that links with underscores (`http://example.org/this_is_a_test`) get properly linkified.  (Note that, while underscores are invalid in host names, they are valid in the rest of a URL, including path, hash, and query.)

The approach aims to be highly efficient by running `linkify` once for each text token (as before), caching the positions of link matches, and applying the linkify rendering if no other rule ends up processing the link in some other way.  Thus in `[link](http://example.org)`, we will detect the URL, but it ends up getting processed by `rules_inline/link.js` before `rules_inline/linkify.js` gets to it.  I did have to modify `rules_inline/text.js` very slightly to stop when it detects a link position, but I believe that's the only impact outside of `linkify.js` code.

I also added a general "preprocess" infrastructure for the inline parser, which is where `linkify` gets called.  There's only one (so far), so this could also be simplified by directly calling `linkify` in `inline_parser.js`'s `parse` function.

One issue [brought up in #38](https://github.com/markdown-it/markdown-it/issues/38#issuecomment-75434351) is handling of `_http://example.com_` and `*http://example.com/*`. Before this PR, both parse to `<p><em><a href="http://example.com/">http://example.com/</a></em></p>`.  With this PR, `_http://example.com_` parses to `<p><em>http://example.com/</em></p>` (because linkify doesn't detect this as a URL), while `*http://example.com/*` parses to `<p>*<a href="http://example.com/*">http://example.com/*</a></p>` (because linkify views the last `*` as part of the URL).  On the other hand, `_<http://example.com/>_` and `*<http://example.com/>*` still parse to the intended `<p><em><a href="http://example.com/">http://example.com/</a></em></p>`.  So this is a possible regression of this PR (though all actual tests still pass).  I believe I can tweak the code (possibly including linkify) to correctly handle these cases if desired (maybe a good idea, but I thought I'd keep it simple for now so the approach is clear).

P.S. I couldn't run `benchmark/benchmark.js` which complained `Cannot find module '../../extra/lib/node_modules/commonmark'`.  Is there some local legwork I need to do to get it running?  I don't have an `extra` directory.


puzrin:
I 100% agree linkifier should be a part of inline parse process, but some details of implementation looks suspicious. For example, [here](https://github.com/markdown-it/markdown-it/pull/580/files#diff-a8851e24cda48d6a505aaa61bb23fa74R49) text plugin become tigh-coupled with links, that's certainly not what it was intended to do.

Need some time to analyze your ideas and remember my postponed proposals, prior to post summary. I general - i'm very interested to accept good implementation.

> P.S. I couldn't run `benchmark/benchmark.js` which complained `Cannot find module '../../extra/lib/node_modules/commonmark'`. Is there some local legwork I need to do to get it running? I don't have an `extra` directory.

https://github.com/markdown-it/markdown-it/blob/master/Makefile#L97

run `make benchmark-deps`. It would be too strange add other parsers as dev deps :). Probably, need add testing directly into benchmark and show more descriptive errors.

There could be regressions, in node 12. For sure - test in node 10 too.

edemaine:
Thanks for the encouraging response! My thought was that having some candidate code would help focus the discussion, but definitely think it'd be good to discuss more.

I agree it's not great that the text plugin needs to be tweaked at all, but I'm also not sure how to avoid it. My main goal was to avoid "dealing with `h` specially", but I still need to check for these special/critical points inside the text plugin, so it's not as elegant as I'd like.  Maybe some kind of dispatch system that plugins can register to be called back when we reach critical positions? But somehow we need to specify that text is "interruptable" by critical points, whereas e.g. links are not interruptable so you don't stop after `(` of `[this](http://example.com)`. Currently this is done by the text plugin code tweak.

Maybe you'd be happier if instead of a `links` list, it's a more general list of "interruption points" for particular positions in a text token that stop the text plugin and give the opportunity for other plugins to look at it?

Thanks for the benchmark pointers. (It might be nice to add `make benchmark-deps` to `CONTRIBUTING.md`, which is where I learned of `benchmark` in the first place. Let me know if you want a PR to that effect.)  Unfortunately, even my small change to the text plugin seems to have cost a lot: ~5% on "current-commonmark" and ~75% on "current" (I guess this enables linkify). ☹️

puzrin:
> I agree it's not great that the text plugin needs to be tweaked at all, but I'm also not sure how to avoid it. My main goal was to avoid "dealing with h specially", but I still need to check for these special/critical points inside the text plugin, so it's not as elegant as I'd like. Maybe some kind of dispatch system that plugins can register to be called back when we reach critical positions?

As far as i remember, my initial idea was to stop on `:`, then parse back to search schema prefix. Because linkifier can be extended with `skype:` and so on, and we still need to fetch it somehow.

> Maybe you'd be happier if instead of a `links` list, it's a more general list of "interruption points" for particular positions in a text token that stop the text plugin and give the opportunity for other plugins to look at it?

Yes, thinked about this. We will invent something more generic, if required.

> Unfortunately, even my small change to the text plugin seems to have cost a lot: ~5% on "current-commonmark" and ~75% on "current" (I guess this enables linkify).

I guess, deopts due access to not existing properties. Try this things:

- always init `state.links = {}` (no type change)
- check via `state.links.hasProperty(pos)` or `Object.prototype.hasProperty(state.links, pos)` or `typeof state.links[pos] !== 'undefined'`

Not sure that helps.

Give me some time, need to understand details for been more certain.

rlidwka:
> Unfortunately, even my small change to the text plugin seems to have cost a lot: ~5% on "current-commonmark" and ~75% on "current" (I guess this enables linkify).

5% may be because of a text plugin, see the advice above about hasProperty checks to see if it can be optimized.

But 75% is another issue entirely. I'm assuming you're looking at the first benchmark based on README.md. It has a lot of links like this:

```md
[![Build Status](https://img.shields.io/travis/markdown-it/markdown-it/master.svg?style=flat)](https://travis-ci.org/markdown-it/markdown-it)
```

Those are markdown links, they don't need to be linkified. But with your solution they do because you need to do pre-processing in order to set up critical points in text.

Essentially, current solution linkified only text, your solution linkifies everything. And since linkifier itself is slow as heck, that's what is tanking benchmarks here IMHO.

puzrin:
@edemaine

While i can understand proposed approach and can explain how to improve speed of extended text breakpoints, suggested principle seems fragile. It tries to detect links in RAW stream, without taking into account possible markup AND possible markup extensions via plugins. 

IMHO it would be more safe to use different and a bit hacky method:

1. Stop on `@` (for emails) and `:` (for links and other schemas).
2. Check if previous text can match email name OR schema (at this point we know available prefix).
3. If prefix matched, verify tails.

Of cause, that will require to extend linkifier API (to allow more fractional use), but that will be less conflicting. Advantages:

1. More predictable result. Less probability of collisions in future - that's very important for maintenance.
2. No need to extend `text` rule scanner (critical chain of parse process)
3. No speed loss - only local ranges of text need to be parsed. I guess, in real world speed may become faster than now.

Let me know what do you think, and are you interested to implement something as above. Or, you may try to explain why your approach should provide better "value" (may be i missed something).

edemaine:
@puzrin Good points -- I agree. The retroactive-at-colon approach seems like a good one. A couple of questions/issues:
1. If we want to catch (currently supported, I believe) schemeless URLs like `//example.com`, we will need to stop on `/` in addition to `:` (for e.g. `mailto:` links) and `@` (for email addresses).  This will require adding `/` to the `isTerminatorChar` list in `rules_inline/text.js`, but hopefully that's a small hit.  `:` and `@` are already on the list, so plugins can already take over at those points.
2. And if we want fuzzy link support too, we'll need to stop at `.` I believe (for e.g. `example.com`).  This is another new terminator character.
3. If you have specific ideas for the linkify-it/linkifier API extensions, I'd be all ears. An obvious API would be `matchAt` to look for a matching URL starting at a specific position (similar in spirit to `testSchemaAt`).  This would mean markdown-it doing some basic left-looking to find the previous word (roughly) and then giving to linkify-it for possible matching.  Alternatively, maybe it would be better to give linkify-it the left-looking task in addition to the matching task, like `findProtocolEndingAt` (for `:`) and `findUsernameEndingAt` (for `@`) and `findHostnameEndingAt` (for `.` fuzzy links) [no left-looking task to do for `//`].  Or maybe these could all be coalesced into something like `findFromPunctuation` or even `matchAtPunctuation` where the index given is the first punctuation of a URL.  Thoughts?

I'm willing to give implementing this a try, unless of course you're interested.  (I'll have to learn the linkify-it code, but hopefully it's simpler than markdown-it 😃.)  Then I'd probably start a new PR on both markdown-it and linkify-it (and close this one).

puzrin:
Hm... you observations are right, i missed some important details in generic description.

## Links

**//example.com**

Seems doable and not need comments. Don't remember how linkify handle this now.

**[fuzzy] emails**

`someone@example.com`

Really useful. Seems covered with described approach and need no comments.

**[fuzzy] links**

`www.google.com`

To be honest, that's a very ugly thing with increased probability of false positive. It's disabled by default and IMO should die :). BUT some people may wish to autolink `www.***` links for legacy reasons (github does that).

I think, we should check that `www.*` links can be supported and announce breaking change for the rest. Or we can suggest old scanning mode (second pass) for strange people with enabled fuzzy links.

## Linkifier API

Regarding to your questions, JS is not very good to work with strings "from position XX". It would be better to split and look into direction of something like `matchLinkHead()` and `matchLinkTail`.

puzrin:
Close - too long without updates. Feel free to create new PR/issue anytime.

PR #596, opened 2019-09-11
Change escapeHtml to escape ' (single quote) and ` (backtick)

co3k:
- HTML5 has Single-quoted attribute value syntax. markdown-it shoud be escape ' avoid XSS on the syntax
- IE recognizes backtick as a valid attribute delimiter and it causes mXSS attacks (https://cure53.de/fp170.pdf). This is not essential safeguard but preferred one

### DIFF
--- a/lib/common/utils.js
+++ b/lib/common/utils.js
@@ -126,7 +126,9 @@ var HTML_REPLACEMENTS = {
   '&': '&amp;',
   '<': '&lt;',
   '>': '&gt;',
-  '"': '&quot;'
+  '"': '&quot;',
+  '\'': '&#039;',
+  '`': '&#096;'
 };
 
 function replaceUnsafeChar(ch) {
--- a/lib/common/utils.js
+++ b/lib/common/utils.js
@@ -127,8 +127,7 @@ var HTML_REPLACEMENTS = {
   '<': '&lt;',
   '>': '&gt;',
   '"': '&quot;',
-  '\'': '&#039;',
-  '`': '&#096;'
+  '\'': '&#039;'
 };
 
 function replaceUnsafeChar(ch) {


puzrin:
I'd like to get more detailed example, how to generate potentially mXSS code via markdown-it (without enabled html option).

co3k:
I think that about mXSS is not an important (just an example) but recognizing \` as attribute delimiter by IE is a bit important.

Opinions of \` might be divided so any comments about it are welcomed.

puzrin:
markdown-it output is restricted and can not produce "anything you wish" if you not enable html (with enabled html output is unsafe and external sanitizer needed). I don't decline discussion, but need narrow down global case to markdown-it specific.

For example:

- markdown-it does not produce `style` attr at all, and escaping `\` does not look necessary.
- attrs are always double quoted, and escaping `'` does not looks necessary.
- untrusted attrs are `src` (protocol-restricted and percent-encoded) and image's `alt` - here backtick can exists.

[Example of backtick in alt](https://markdown-it.github.io/#md3=%7B%22source%22%3A%22!%5B%5C%5C%60test%5D%28%60test%29%5Cn%22%2C%22defaults%22%3A%7B%22html%22%3Afalse%2C%22xhtmlOut%22%3Afalse%2C%22breaks%22%3Afalse%2C%22langPrefix%22%3A%22language-%22%2C%22linkify%22%3Atrue%2C%22typographer%22%3Atrue%2C%22_highlight%22%3Atrue%2C%22_strict%22%3Afalse%2C%22_view%22%3A%22src%22%7D%7D)

So, my question is - which cases exactly to you try to avoid, browser versions and so on.

Side note: even if you need advanced escaping for attrs, it should not pollute [normal text](https://github.com/markdown-it/markdown-it/blob/master/lib/renderer.js#L113)

PS. Independent on this issue result, it was useful for me to read referred document, thank you.

co3k:
> attrs are always double quoted, and escaping ' does not looks necessary.

In a core implementation of markdown-it, it should be true, but in some markdown-it plugins might not be true.

We can't restrict using single quoted attr in markdown-it plugins. As defined in HTML5, there should be legitimacy to use it. In this case, it is equally conceivable to use the utility function exported by markdown-it to perform HTML escaping. In fact, it's a great way to avoid making one's own.

I don't have a strong opinion about backtick, but I insist that escaping single quotes is clearly necessary.

puzrin:
> In a core implementation of markdown-it, it should be true, but in some markdown-it plugins might not be true.

That's not concrete (too generic). Please, provide concrete samples for EVERY additional char, suggested for escape. That's very important for moving forward.

co3k:
... okay, I will drop escaping backtick. How about this?

co3k:
Escaping backtick is not special but I also think it is not an important so dropping backtick support is good for "moving forward".

https://github.com/mathiasbynens/he/blob/master/he.js#L49
https://github.com/wooorm/stringify-entities/blob/master/index.js#L19

puzrin:
Close - doesn't seems to be actual. Could no get proofs about anything not <= IE8.

PR #599, opened 2019-10-02
Add support to use plugins on markdown-it CLI with --plugin parameter

pnizer:
With this change, it's possible to add plugins on makrdown-it CLI execution. For exemple, to use the markdown-it-mermaid and markdown-it-footnote plugins, we can just call the markdown-it command with --plugin argument:
```
cat markdown.md | markdown-it --plugin markdown-it-mermaid --plugin markdown-it-footnote
```

### DIFF
--- a/bin/markdown-it.js
+++ b/bin/markdown-it.js
@@ -47,6 +47,11 @@ cli.addArgument([ '-o', '--output' ], {
   defaultValue: '-'
 });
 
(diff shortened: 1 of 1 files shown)


puzrin:
That will introduce new problem with missed plugin options support. I don't like such incomplete feature.

Also, no tests and broken travis-ci.

puzrin:
Close - incomplete/broken

PR #605, opened 2019-10-31
No longer replace two periods with ellipsis

ptetau:
Two periods i.e. ".." are often used in representing ranges, e.g. 1..99, means a sequence of numbers 1, 2, 3, 4 all the way to 99 (see F# ranges). The current markdown-it replacements currently clobber the correct sequence with an invalid syntax. 
This change allows the replacement for 3 or more periods, instead of the current 2 or more.

### DIFF
--- a/lib/rules_core/replacements.js
+++ b/lib/rules_core/replacements.js
@@ -63,9 +63,9 @@ function replace_rare(inlineTokens) {
       if (RARE_RE.test(token.content)) {
         token.content = token.content
           .replace(/\+-/g, '±')
-          // .., ..., ....... -> …
+          // ..., ....... -> …
           // but ?..... & !..... -> ?.. & !..
-          .replace(/\.{2,}/g, '…').replace(/([?!])…/g, '$1..')
+          .replace(/\.{3,}/g, '…').replace(/([?!])…/g, '$1..')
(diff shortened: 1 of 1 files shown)


puzrin:
It's not good idea to break other cases in preference of your own. You can do all required modifications locally.

ptetau:
> It's not good idea to break other cases in preference of your own. You can do all required modifications locally.

However this is an unexpected replacement, even the existing commented documentation suggests it only replaces 3 or more periods. 
> // ... → … (also ?.... → ?.., !.... → !..)

puzrin:
This IS expected replacement, covered by tests, been broken by your change.

ptetau:
I have fixed the test to cover the correct behaviour, and I can make another pull request. I can't find any precedence for a two period ellipsis which is why I think this behaviour is wrong. The definitions are rather precise regarding 3 periods:
1. https://en.wikipedia.org/wiki/Ellipsis
2. https://getitwriteonline.com/articles/ellipses
3. https://www.grammarly.com/blog/ellipsis/
4. https://www.grammarbook.com/punctuation/ellipses.asp

The typographic feature in this case is being overzealous, acting as a spellcheck.
There is some irony here as ellipsis indicates omission and a two period ellipsis, is omitting the final period.

Issue #584, opened 2019-08-14
where is all the built-in rules?


zzzgit:
Hi, I want to disable some of the built-in rules.

Where can find all the built-in rule names?
I failed to find it in the documentation.

puzrin:
You need to look into sources:

- https://github.com/markdown-it/markdown-it/blob/master/lib/parser_inline.js
- https://github.com/markdown-it/markdown-it/blob/master/lib/parser_block.js
- https://github.com/markdown-it/markdown-it/blob/master/lib/parser_core.js

DimaDK24:
I want to do the same. Perhaps it's worth mentioning at least links above somewhere in docs?

puzrin:
That's opinion-based. You can suggest PR with improvements to discuss.

PR #609, opened 2019-11-03
Add links to all default rules to README

DimaDK24:
There's a function to disable any rules, but all default rules are not described anywhere in docs. This PR adds links to rules in sources to README

relates to #584


puzrin:
Thanks

PR #613, opened 2019-11-07
Add ESM support

lpellegr:
Browsers have improved a lot over the past years and it is now possible to do modern web development without requiring any build tools (i.e. buildless) with ESM support.

We are adding ESM support to Prose Mirror and as a requirement, transitive dependencies need to provide an ESM version: 

https://discuss.prosemirror.net/t/prose-mirror-buildless-esm-support/2322

Markdown-it is one of these dependencies.

https://github.com/ProseMirror/prosemirror-markdown

This PR aims to generate an ESM version of the source code before packaging.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@ dist/
 node_modules
 support/demo_template/sample.js
 benchmark/extra/
+rollup.config.js
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ coverage/
 demo/
 apidoc/
 *.log
+*.mjs
--- a/package.json
+++ b/package.json
@@ -12,14 +12,18 @@
   "repository": "markdown-it/markdown-it",
   "license": "MIT",
   "main": "index.js",
+  "module": "dist/markdown-it.mjs",
   "bin": {
     "markdown-it": "bin/markdown-it.js"
   },
   "scripts": {
+    "build": "rollup -c",
+    "prepare": "npm run build",
     "test": "make test"
   },
   "files": [
     "index.js",
+    "index.mjs",
     "bin/",
     "lib/",
     "dist/"
@@ -56,6 +60,10 @@
     "mocha": "^6.1.4",
     "ndoc": "^5.0.0",
     "pug-cli": "^1.0.0-alpha6",
+    "rollup": "^1.26.3",
+    "rollup-plugin-commonjs": "^10.1.0",
+    "rollup-plugin-json": "^4.0.0",
+    "rollup-plugin-node-resolve": "^5.2.0",
     "stylus": "^0.54.5",
     "supertest": "^4.0.2",
     "terser": "^4.1.2"
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,0 +1,18 @@
+import commonjs from 'rollup-plugin-commonjs';
+import nodeResolve from 'rollup-plugin-node-resolve';
+import json from 'rollup-plugin-json';
+
+export default {
+  input: 'index.js',
+  output: {
+    file: 'dist/markdown-it.mjs',
+    format: 'esm'
+  },
+  plugins: [
+    json(),
+    nodeResolve({
+      preferBuiltins: false
+    }),
+    commonjs()
+  ]
+};


lpellegr:
Do you think this could be merged and released after a review?

lpellegr:
@puzrin ping

puzrin:
I don't like such change, because it adds one more build step (+ dist for browsers) and does not provides expected benefits of esm modules for internals of markdown-it.

It would be more logical switch to esm as main format and leave cjs as legacy (and replace browser build with that cjs), but right now node supports `.mjs` only behind flags. https://github.com/nodejs/modules#current-efforts node team plans to remove that flag soon. I think, it would be more logical to wait until that happens + 1-2 month for spread, and try use "native" solution.

lpellegr:
@puzrin Thanks for your feedback.

I understand your point of view but all major browsers are supporting ESM imports since months. As you said, node support is still experimental. Switching the codebase to ESM is the best solution but it will take weeks to months. Besides, there is already a large ecosystem that supports the non-standardized but common _module_ field in `package.json`.

What about this PR as transition? it introduces no breaking changes. Once you will have transitioned to ESM as the main format, you can update the packaging with non-breaking changes again for users. 

Regarding the build step, it is minimal and takes less than 1 second to complete. About the package size, the increase will be similar once you switch to ESM as main and provide a CJS version aside.

PS: I have updated the PR to put the generated _mjs_ file in `dist/`.

puzrin:
> What about this PR as transition?

Changes in development process are very expensive, because should spread to multiple repos, supported by me. I have absolutely no motivation to spend my time for doing this twice.

lpellegr:
Why this change should spread to multiple repos?

puzrin:
Because i don't like to have custom support requirements in each.

lpellegr:
I don't understand. The addition that is discussed there is beneficial to users who need to have a manner to consume _markdown-it_ with no build steps on modern web apps (especially during dev). Since this is no breaking change, there is absolutely no need to update/consume the change if you don't require it. It's not a custom requirement but a general change for a usage that is more and more frequent.

puzrin:
1. I'm not against esm support. I don't like concrete implemenattion. Would prefer native one.
2. It's not rationale to make such changes twice, if node will have esm support enabled very soon.

lpellegr:
1. I agree with you but nodejs support is experimental and once released do you think the existing web ecosystem will have support for this new method from day one?  
2. You mean taking 10 minutes to merge a 28 lines patch and performing a release?

puzrin:
> 1. I agree with you but nodejs support is experimental and once released do you think the existing web ecosystem will have support for this new method from day one?  

Have you seen link i provided? They have schedule https://github.com/nodejs/modules#current-efforts. This months + 1-2 months to polish. I don't know what you mean about web ecosystem and how that is related to esm support.

> 2. You mean taking 10 minutes to merge a 28 lines patch and performing a release?

I mean maintenance cost in long term and risks of development/release process change. It seems, we start falling into generic development questions, not specific to this package. I'd like to avoid spending time for such talks.

lpellegr:
By ecosystem, I mean web tools that already support the _module_ field.

> I mean maintenance cost in long term and risks of development/release process change. It seems, we start falling into generic development questions, not specific to this package. I'd like to avoid spending time for such talks.

It's not as different as providing a minified version of the library. Why are doing it? 

As briefly explained previously, the maintenance cost is close to zero since the file is auto-generated. In the case of process change, I doubt that 28 lines of non-breaking code (with only 3 that may affect the process) are a problem...

Anyway, there are people who are reluctant to changes and I respect that but when there are no real reasons it's a shame.

puzrin:
In my eyes that looks different. Support of esm can be added in prefered way ~ after 3 months. But you don't care about my time and want force me do this twice, using not constructive and not polite comments. Please, stop it. If you have default permission to post here, that does not mean everything you can post is welcome. If you think, you understand in development better than me - do a fork and don't waste my time.

patrickkettner:
Hi @puzrin!
As 13.2, [which removed the flag](https://medium.com/@nodejs/announcing-core-node-js-support-for-ecmascript-modules-c5d6dc29b663), was released a few months ago, but as mentioned in the blogpost "the implementation remains experimental and subject to change"

would you be open a PR that has a native implementation? as you described [earlier](https://github.com/markdown-it/markdown-it/pull/613#issuecomment-551367665), or do you want to wait until it is no longer listed as experimental?

puzrin:
@patrickkettner see https://github.com/nodeca/js-yaml/issues/541 with similar question. If anyone can explain all details of new & comfortable develompent process, i'm ok to migrate. Using 13.2 is accepatble.

Issue #617, opened 2019-12-05
Custom rules documentation


rl-saen:
Hello everyone,
my CSS dude wants me to add a wrapper element within every `li` when translating markdown to HTML.

Looking at #598 I see it is possible to inject custom rules and functionality into `markdown-it` by
modifying `mdi.renderer.rules`. However, I can't find documentation in the readme as to which keys `rules` has/expects for certain elements. 

When displaying the content of `rules` at build time it looks like only default rules are present, meaning I have to add new rules myself. But again, I am not sure how to properly extend `rules`: Looking at #598 I see the function itself doesn't differentiate between elements it applies its changes to all values it receives. So I guess it depends on the key name in `rules` which element the function is applied on? If so where can I find a list of which key names are available?

Best regards

puzrin:
Advanced modifications require some sources instept. See https://github.com/markdown-it/markdown-it/blob/master/lib/renderer.js

It uses default renderer function if "named" one not exists. Seems simple.

rl-saen:
> It uses default renderer function if "named" one not exists. Seems simple.

Indeed, but to know which rule I'd have to create I have to know what to name my key in the rule object.

Again looking at #598 I see that to add a rule to access the opening tag of a link I have to set the function at the key `link_open`. My question in detail is "What is the correct key to add to the rules object to access the opening tag of a `li` Element?" and in general "Is there a full list somewhere of the possible key names for the `rules` object?"

> Advanced modifications require some sources inspect.

Yes, I did  read through parts of the source code but didn't find a solution (maybe looked in the wrong place, etc etc.). 

I find it especially strange that the readme explicitly mentions (in the 6th line none the least)
> Configurable syntax! You can add new rules and even replace existing ones.

But then there is no follow up on how to do that.

puzrin:
Parser is non trivial thing. We have no resources to write a big book. If you describe exactly what you need, i could try to help with direction.

rl-saen:
>What is the correct key to add to the rules object to access the opening tag of a li Element?

and
>Is there a list somewhere of the possible key names for the rules object?

puzrin:
[...]

PR #619, opened 2019-12-07
Added examples on how to add and modify rules

saenglert:
Hello everyone,
thanks again @puzrin, for helping me out in #617 (@rl-saen is the GitHub account I use at work).

You said you don't have the time write a book about how to use markdown-it, so I thought I'd help out a little and write a guide on the problem you helped me solve. It is 75% finished, only for the last example I need a little help on how to properly create and insert a new token. 

Can you please take a look at it?

### DIFF
--- a/support/demo_template/index.pug
+++ b/support/demo_template/index.pug
@@ -65,7 +65,7 @@ html
             input#linkify(type='checkbox')
             |  linkify
         .checkbox.not-strict
-          label._tip(title='do typographyc replacements, (c) -> © and so on')
+          label._tip(title='do typographic replacements, (c) → © and so on')
             input#typographer(type='checkbox')
             |  typographer
         .checkbox.not-strict


rlidwka:
> Can you please take a look at it?

Okay, I've taken a look at it. Please rename the file to `renderer rules` (see my explanation in the review above), and we can merge it hoping that it may help someone.

But honestly, I've never felt this part (renderer) needs any documentation. It's intended as a simple way for people to add custom hacks.

----

There is another part of markdown that needs documentation more urgently. Plugins that work before renderer:

```js
function add_span_plugin(md) {
  md.core.ruler.push('ul_add_custom_span', function add_span(state) {
    let new_tokens = []
    let inside_list = false

    for (let token of state.tokens) {
      if (token.type === 'list_item_open') inside_list = true
      if (token.type === 'list_item_close') inside_list = false
      if (token.type === 'inline') {
        token.children.forEach(token => token.level++)
        token.children.unshift(new state.Token('span_open', 'span', 1))
        token.children.push(new state.Token('span_open', 'span', -1))
      }
    }
  })
}

let mit = require('./')
let md = new mit().use(add_span_plugin)

console.log(md.render('- hello world'))
// <ul>
// <li><span>hello world</span></li>
// </ul>
```

This is a way to add your example `span` while remaining compatible with any other plugin that potentially wants to do the same thing.

And it's better in every way except... there is no documentation for this, and there's no way people can find out how to write these plugins. :(

saenglert:
Wow, almost forgot this PR still exists!

I corrected the spelling mistakes, added some more context thanks to your explanation and fixed the last code example.

And yeah, I agree your middleware solution seems a lot more streamlined. As my focus at work has shifted quite a bit I am currently not using markdown-it anymore and even had some trouble getting the last example to work properly as I am not really used to it anymore.

So I hope someone else can step into the breach and pick this up.

rlidwka:
Okay, merged. This is basic stuff, but I hope it will still be useful.

Unfortunately, we don't have time and resources to maintain such documentation. So it may get partially or fully removed if examples stops working due to breaking changes in the future.

If someone wants to expand on this or add more documentation, maybe we should start a wiki.

PR #634, opened 2020-01-24
Update blockquote.js

iMrDJAi:
Added a spoiler tags syntex check `>!spoiler!<`, so it will return false if `>` is followed by `!`.


puzrin:
Sorry, we do not accept changes, not related to commonmark spec. Policy is to provide flexibility, instead of incorporate all possible extensions users can imagine.

If you wish to share your work, you can:

- Publish your plugin as separate package.
- Promote desired blockquote change at commonmark forum. If it will be accepted, spread here after some time.

iMrDJAi:
OK, I understand.

Issue #598, opened 2019-09-12
Create attribute with null value


networkimprov:
Can extensions generate an empty attribute, e.g. `<a href="..." download>...</a>`?
If not, could that be enabled?

I tried these:
```
mdi.renderer.rules.link_open = function(iTokens, iIdx, iOptions, iEnv, iSelf) {
   iTokens[iIdx].attrs.push(['download', null]); // yields download="null"
   //                       ['download']            yields download="undefined"

   return iSelf.renderToken(iTokens, iIdx, iOptions);
};
```

puzrin:
https://github.com/markdown-it/markdown-it/blob/master/lib/renderer.js#L177

You may override attrs renderer to support empty values.

networkimprov:
Understood, thanks.

Could you consider making a `null` attrs value yield an empty HTML value in 11.x?

puzrin:
That's strange - nobody asked such feature for 5 years.  But i have no principal objections.

rlidwka:
> Could you consider making a `null` attrs value yield an empty HTML value in 11.x?

Alternative suggestion would be to represent empty values as `attrs.push(['download', true])`.

I think many people would expect `attrSet('download', undefined)` to *unset* the existing attribute rather than set it to empty value.

puzrin:
I'd suggest use notation `attrSet('download')`, it's clear. Also, i'd avoid use attrSet for unsetting and any kind of unsetting in our api at all.

Lemmingh:
## I don't think any change is required to support boolean attribute

Your "empty attribute" is canonically named "boolean attribute".

The following work:

```javascript
attrSet("download", "")

attrSet("download", "download")
```

## I'm afraid the discussion here reveals a few potential bugs of `markdown-it`

<https://github.com/markdown-it/markdown-it/blob/6e2de08a0b03d3d0dcc524b89710ce05f83a0283/lib/token.js#L170-L181>

`attrGet()` uses `null` to indicate the absence. Thus, both `null` and `undefined` should be reserved. But `attrSet()` and `renderAttrs()` don't validate it:

<https://github.com/markdown-it/markdown-it/blob/6e2de08a0b03d3d0dcc524b89710ce05f83a0283/lib/token.js#L153-L167>

<https://github.com/markdown-it/markdown-it/blob/6e2de08a0b03d3d0dcc524b89710ce05f83a0283/lib/renderer.js#L168-L185>

---

Besides, the access control on `Token` should be reinforced.

[...]

PR #652, opened 2020-03-22
feat: support boolean attributes

privatenumber:
https://github.com/markdown-it/markdown-it/issues/598

Supports:
- `.attrSet('download')` on Token
- `.attrs.push(['download'])` on Token
- `.attrs.push(['download', null])` on Token
- `.attrs.push(['download', undefined])` on Token

### DIFF
--- a/lib/common/utils.js
+++ b/lib/common/utils.js
@@ -2,6 +2,7 @@
 //
 'use strict';
 
+function isNil(v) { return v === null || typeof v === 'undefined'; }
 
 function _class(obj) { return Object.prototype.toString.call(obj); }
 
@@ -299,6 +300,7 @@ exports.lib                 = {};
 exports.lib.mdurl           = require('mdurl');
 exports.lib.ucmicro         = require('uc.micro');
 
+exports.isNil               = isNil;
 exports.assign              = assign;
 exports.isString            = isString;
 exports.has                 = has;
--- a/lib/renderer.js
+++ b/lib/renderer.js
@@ -11,6 +11,7 @@
 var assign          = require('./common/utils').assign;
 var unescapeAll     = require('./common/utils').unescapeAll;
 var escapeHtml      = require('./common/utils').escapeHtml;
+var isNil           = require('./common/utils').isNil;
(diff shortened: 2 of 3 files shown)


lib/common/utils.js
```diff
@@ -299,6 +300,7 @@ exports.lib                 = {};
 exports.lib.mdurl           = require('mdurl');
 exports.lib.ucmicro         = require('uc.micro');
 
+exports.isNil               = isNil;
```
puzrin: It's not good idea to spread micro helpers as lib feature, when used in only one place. Such things clutter api significantly in long term.

privatenumber: Was trying to follow the existing organization.

Where do you want it?

puzrin:
We need do discuss final data format in original issue. I don't like duality (null & undefined) for single case, especially when it's hidden behind api call.

privatenumber:
@puzrin 

I'm okay with just checking against `undefined`, but that would mean that doing `.setAttrs('attr', null)` -> `attr="null"` is expected behavior.

Let me know what you decide and I'll update.

lib/common/utils.js
puzrin: Until "helper" is not used wide, it should be in the same file where it's needed. Preferably - inlined, without separate fn at all.

puzrin:
> I'm okay with just checking against `undefined`, but that would mean that doing `.setAttrs('attr', null)` -> `attr="null"` is expected behavior.
> 
> Let me know what you decide and I'll update.

I understand. I left issue open to discuss signature proposals from other users. Did not thinked about it myself, because issue is not too hot and can be hacked at user side (without dependency on my personal opinion).

septatrix:
I have no strong preference but would like to add that boolean attributes can also be set to their own name to signify a true value (e.g. `<div hidden="hidden">Something hidden and HTML5 compliant</div>`) which is what we just settled with for now.

puzrin:
Closing as stale. Still need to finalize the data format before any code change. See original issue.

PR #653, opened 2020-03-22
refactor: use findIndex

privatenumber:
Refactor to use [`#findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) where appropriate.

### DIFF
--- a/lib/ruler.js
+++ b/lib/ruler.js
@@ -48,12 +48,9 @@ function Ruler() {
 // Find rule index by name
 //
 Ruler.prototype.__find__ = function (name) {
-  for (var i = 0; i < this.__rules__.length; i++) {
-    if (this.__rules__[i].name === name) {
-      return i;
-    }
-  }
-  return -1;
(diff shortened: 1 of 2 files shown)


puzrin:
No refactoring "just for fun" please. Current code is speed optimized. We can not accept aestetic-based changes.

privatenumber:
@puzrin 

Whats the browser/Node support baseline for markdown-it?

privatenumber:
I'm thinking without a support baseline and more concrete benchmarking for each env markdown-it supports, adopting native methods that reduce the distribution size and come with compiler level optimizations will always look like "aesthetic-based changes".

puzrin:
ES5-capabale (including polyfills).

In this case - adding closure functions will not improve features, and 100% will not improve speed. If you need changes - you are expected to provide clear metrics what will be improved and how much. Commits without clear criteria are always declined by default.

PR #665, opened 2020-05-14
Update TravisCI to explicit hidden declarations

edumco:
This PR changes the Travis configuration to make all configurations explicit so other collaborators can easily understand what is really going on.


puzrin:
It's intended choice to keep configs compact, with only valuable data.

PR #682, opened 2020-06-14
README: add import syntax; modernize

dandv:


puzrin:
This package is still compatible with es5. I see no reasons to change readme for obvious things.

Issue #689, opened 2020-07-16
Table layout could be disrupted.


ataihei:
When using table syntax, if you use even one `` ` `` after surrounding a `` ` `` with a ``` `` ```, the table layout may be corrupted.

Example:
```markdown
|a|b|
|--|--|
|`` ` ``| `c` |
|`` ` ``| `d |
| `e` | `` ` `` |
|`` ` ``| f |
```
Result:
![image](https://user-images.githubusercontent.com/10428970/87619130-68590200-c756-11ea-81b7-2abcef51de61.png)

Thank you for your response.

rlidwka:
Fixed, should be the same as on github now.

PR #697, opened 2020-07-31
Fix tables.js for pipes inside backticks

klassiker:
Add test samples and fix wrong test sample

The sample `Pipes inside backticks don't split cells` did not contain the output that github would create from that markdown.

```
 | Heading 1 | Heading 2
 | --------- | ---------
 | Cell 1    | Cell 2
 | `Cell|3`  | Cell 4
```
produces

 | Heading 1 | Heading 2
 | --------- | ---------
 | Cell 1 | Cell 2
 | `Cell|3` | Cell 4

If the current cell is backTicked, the current char is a pipe and the previous char is not an escape, the cell ends.

After adjusting that it is no longer necessary to check if the end has been reached to due an unmatched backtick.

### DIFF
--- a/lib/rules_block/table.js
+++ b/lib/rules_block/table.js
@@ -19,8 +19,7 @@ function escapedSplit(str) {
       ch,
       escapes = 0,
       lastPos = 0,
-      backTicked = false,
-      lastBackTick = 0;
+      backTicked = false;
 
   ch  = str.charCodeAt(pos);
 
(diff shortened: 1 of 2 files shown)


puzrin:
Thanks for PR. All pending issues will be reviewed in 102 weeks. Sorry for delay.

barro350:
I am not sure whether `table.js` should deal with backticks at all.

#86 made backticked pipes in table cells not parsed as delimiters. Back then, it was consistent with GFM. However, after 2 years later, GFM changed their spec; see https://github.com/github/cmark-gfm/issues/24 and dicussion therein. GFM now does parse every `|` as a table delimiter unless it is escaped.

Current implementation of `table.js` has another issue: #689 (e.g., the case in which a table cell contains complicated backtick sequences like ``` `` ` `` ```), which will also be fixed if `table.js` does not deal with backticks at all.

puzrin:
@barro350 thanks for reference. We will sync with spec. Tables are intended be close to github. But implementation was done before spec available.

NOTE. This will be breaking change.

rlidwka:
I've changed `table.js` to be similar to up-to-date gfm:
 - table now doesn't deal with backticks at all
 - `|` is considered escaped if and only if there is a `\` character immediately before it
 - number of elements in the first row (thead) now must match second row (aligns) exactly
 - no tbody if it would be empty

I believe it fixes this issue (all your tests pass), so closing this PR.

Issue #626, opened 2020-01-11
Pass additional context to highlight option from fence renderer


KyleMit:
This is similar to #456 (Access to the render function from a custom highlight function) but much  more limited in scope. 

In some markdown flavors like [python][0], you can pass additional parameters at the head of a fenced code block like this:

     ```python hl_lines="1"
     # This line is emphasized
     # This line isn't
     ```

Currently, the `options.highlight` callback passes two parameters:

```
options.highlight(token.content, langName)
```

Where `langName` is created by taking all the info available at the head of the fenced code block  and splitting on the first whitespace character

```js
var info = unescapeAll(token.info).trim()  // python hl_lines="1"
var langName = info.split(/\s+/g)[0]       // python
```

The only requested delta is to leave the first two parameters exactly as they are and just pass in the rest of the available context as a third so the highlight function can make use of it:

```diff
if (options.highlight) {
-    highlighted = options.highlight(token.content, langName);
+    highlighted = options.highlight(token.content, langName, info);
```

### Possible workarounds

The same workaround to #456 is available here to override the [`default_rules.fence`][1] implementation, but that leaves a lot of work in the client's flow

Another available workaround is to delimit by a non-space character when shoving extra info to fenced code blocks like this `python,hl_lines="1"`, and then splitting out the language definition inside the highlight function, but this breaks the language detection for other engines where the raw md might live, like in your editor or github.

[0]: https://python-markdown.github.io/extensions/fenced_code_blocks/
[1]: https://github.com/markdown-it/markdown-it/blob/10.0.0/lib/renderer.js#L39

puzrin:
Suggestion sounds reasonable, but i'm very conservative about api changes. I see no strong reasons to decline immediately, but i don't like to make this change too :).

I'd stay this open for a while.

KyleMit:
Seems reasonable - i was able to [locally override the fence handler](https://github.com/VermontDepartmentOfHealth/docs/commit/37abb5da#diff-64bbf3709cc6a42c12f4c84016d29a36) in my own project, so that workaround handles my own workflow.  Only needs to be pulled into core if it helps other folks too.  Thanks

rlidwka:
Fixed, remainder of fence params is now passed through to highlighter as 3rd option.

danieleades:
@rlidwka will this fix only work if the additional attributes are whitespace-delimited, or will it also work if they're comma-separated?

As in-

````
```rust,no_run
fn foo(bar: &Bar) {
[...]

PR #706, opened 2020-09-02
Add the additional fence attributes to calls to `highlight` so that highlighters can then add them to the <pre> tag

pragdave:
I need to be able to pass additional parameters to highlights, such as

    ~~~ javascript {line-numbers=5 highlight=14-17}

These attributes were not being forwarded from the default fence renderer to the `highlight` function.

I've added them in as a third parameter, so the change will be transparent to existing code.


Cheers

Dave

### DIFF
--- a/lib/renderer.js
+++ b/lib/renderer.js
@@ -47,7 +47,7 @@ default_rules.fence = function (tokens, idx, options, env, slf) {
   }
 
   if (options.highlight) {
-    highlighted = options.highlight(token.content, langName) || escapeHtml(token.content);
+    highlighted = options.highlight(token.content, langName, token.attrs) || escapeHtml(token.content);
   } else {
     highlighted = escapeHtml(token.content);
   }


puzrin:
~ dupe of #626

Could you post there your use cases? That will help to verify possible signatures.

pragdave:
Here's a typical case:

~~~~
~~~ javascript {highlight-a="1,4,6,9" highlight-b="2,3"}
function codeFor(name) {
  for (let i = 0; i < CountryCodes.length; i++) {
    if (name == CountryCodes[i][0])
      return CountryCodes[i][1]
  }
  return null
}

codeFor("Canada")
~~~
~~~~

My `highlighter()` looks for attributes starting `highlight-` and adds `<mark>` tags to the given lines, where the class of the tag depends on the attribute name.

So there are no fixed attribute names: this is just a way of passing a set of values from the markdown level to the stuff that does rendering.

In my code, the type is simply `Record<string,string>[]`.

lib/renderer.js
```diff
@@ -47,7 +47,7 @@ default_rules.fence = function (tokens, idx, options, env, slf) {
   }
 
   if (options.highlight) {
-    highlighted = options.highlight(token.content, langName) || escapeHtml(token.content);
+    highlighted = options.highlight(token.content, langName, token.attrs) || escapeHtml(token.content);
```
rlidwka, line 50: Where do you get `token.attrs` from? Shouldn't it be second half of `token.info`?

pragdave:
Oh, I'm sorry. I had the `-attrs` plugin loaded, and it added `attrs` the token, which I just passed on.

Hmmm... I really would like to see this functionality: I'm using it daily. Would you like me to

1. Add attribute parsing here, or
2. Just pass a string through to the highlighter function?

Dave

rlidwka:
Added remainder of fence params as 3rd argument to highlighter, I hope it fixes this issue.

pragdave:
Thanks!

PR #745, opened 2020-12-20
Deep copy `token.attrs` to avoid mutating token stream

hukkinj1:
Please correct me if I'm wrong but I believe this line https://github.com/markdown-it/markdown-it/blob/be22253b7ccb98ff69f010935a034d601d80ceec/lib/renderer.js#L72 will currently mutate the token stream.

This PR makes a deep copy of `token.attrs` so that mutating `tmpAttrs[i]` will not mutate `token.attrs[i]`


puzrin:
Thanks! I did it a bit more light way.

PR #747, opened 2021-01-02
Instantiating linkify-it lazily, only if needed

fabiospampinato:
This PR enables instantiating linkify-it lazily, meaning never if it's disabled.

The advantages are, when linkify-it is disabled: 

- shaving ~4ms (~20%) from startup time (~1.5ms for requiring the module and ~2.5ms for instantiating it), which is relatively marginal but it adds up.
- reducing memory usage by half a megabyte, or more if more than one instantiation of linkify-it is prevented.
  - before:
    image
  - after
  	image

I know adding more code and making the index file's code uglier is something that the maintainers here want to prevent as much as possible, so I've tried to make the diff as clean as possible. Hopefully this is clean enough.

### DIFF
--- a/lib/common/utils.js
+++ b/lib/common/utils.js
@@ -39,6 +39,24 @@ function arrayReplaceAt(src, pos, newElements) {
   return [].concat(src.slice(0, pos), newElements, src.slice(pos + 1));
 }
 
+// Sets a property whose initial is computed lazily on an target object
+function defineLazyProperty(target, property, instantiator) {
+  var initied = false;
+  var value;
+  Object.defineProperty(target, property, {
+    get: function () {
+      if (initied) return value;
+      initied = true;
+      value = instantiator();
+      return value;
+    },
+    set: function (newValue) {
+      initied = true;
+      value = newValue;
+    }
+  });
+}
+
 ////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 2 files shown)


puzrin:
Seems someone already tried to suggested something abou lazy init. Answer was "no really useful added value".

Do you have real use cases where suggested improvement makes sense?

fabiospampinato:
Yeah that was me. I don't really have a use case where ~4ms or 500KB on their own make a night and day difference.

Feel free to close this if you don't think it's worth it.

puzrin:
I'd prefer to avoid hacks wihout need. Let's close until use case become available.

Also, probably worth to optimize linkify-it instead (but only after use case available)

animeshsa:
I have faced the same problem. I used in a for loop and saw in memory tab string object linkify took hugely memory. I need this patch. please help

animeshsa:
> I'd prefer to avoid hacks without need. Let's close until use case become available.
> 
> Also, probably worth to optimize linkify-it directly instead (but only after use case available)

I have faced the same problem. I used in a for loop and saw in memory tab string object linkify took hugely memory. I need this patch. please help

puzrin:
GC logic depends on available memory. If you have a lot of free memory, JIT can eat very much. "for loop" observations are naive and can not be considered as metric of memory consumption.

PR #748, opened 2021-01-03
Various performance improvements

fabiospampinato:
I tried to optimize the library a bit, there wasn't much that could have been easily optimized away but there was something.

I don't expect this PR to get merged in its entirety, or at all, I'm mostly writing this as a brain dump, but there are actually a few very minor changes I found that make some signficant difference, maybe those should be ported over here, your choice.

##### Summary

Overall the profile script now runs in ~23% less time.

All tests conducted under Node v15.5.0, older versions might give considerably different results.

<details>
  <summary>10 sample runs (Before)</summary>

```
profile: 2.516s
profile: 2.537s
profile: 2.695s
profile: 2.658s
profile: 2.508s
profile: 2.537s
profile: 2.572s
profile: 2.526s
profile: 2.536s
profile: 2.587s

avg: 2.567s
improvement: 0%
```

</details>

<details>
  <summary>10 sample runs (After)</summary>

```
profile: 1.965s
profile: 1.940s
profile: 1.924s
profile: 1.920s
profile: 2.085s
profile: 2.063s
profile: 1.929s
profile: 1.948s
profile: 1.935s
profile: 1.979s

avg: 1.969s
improvement: 23.3%
```

</details>

##### Optimizations

- Significant:
  - `lib/rules_core/normalize.js`
    - optimizing this single rule gave by far the largest performance gains of the lot. 
    - turns out that replacing "\n" with "\n" is a significant performance burden, so just removing "|\n" from the regex, a 3 characters diff, would be a significant performance improvement.
    - because of a [bug](https://bugs.chromium.org/p/v8/issues/detail?id=10737&sort=-stars&q=&can=4) in v8 looking for "\0" in strings that don't contain only ASCII character is a significant performance cliff, just iterating over the entire thing and performing the change manually turned out to be significantly faster.
      - by the way I don't see the point of replacing "\0" in JS so this replacement should probably be removed entirely, breaking away slightly from CM as this requirement by the spec has only downsides in JS really. 
- Minor:
  - `lib/common/utils.js`
    - Using `String.prototype.repeat` whenever possible, that's 10x faster than the previous implementation and more memory efficient.
  - `lib/renderer.js`
    - `fence`
      - most of the times the "info" string doesn't contain any spaces, so I added a fast path for that that doesn't do so much wasteful stuff.
      - asking for the index of `<pre` is not actually what we need, we don't want to scan the entire thing, we only care about how it starts, so it now only considers the first 4 characters of the string.
      - when adding the "class" attribute pretty much always there's no need to clone the current attributes array at all, so I added a more optimized path for that common case.
  - `lib/rules_block/state_block.js`
    - apparently creating an intermediary array is slower than just progressively joining those lines (~90ms vs ~15ms).
- Very Minor:
  - some unnecessary property accesses have been removed.
  - some exit conditions in rules have been moved slightly up in the functions, ensuring those functions exit as quickly as possible when those exit conditions are met.
    - this was a pretty noisy change with not that much to gain from really.

##### Other Changes

- `benchmark/benchloop.js`
  - I added another profiling script, one that's not so tied to how the CM spec is written and that gives some output quickly.
- `benchmark/profile.js`
  - 10x more iterations are executed, giving more time to v8 to optimize things.
  - A number is outputted, otherwise this file is not very useful IMO, timing the entire script is not what we want as that would count disk reads etc.

##### Potential Future Optimizations

- Near future
  - something like 30% of the time spent is now spent html-escaping strings, maybe that could be optimized somehow, everything I tried so far didn't work. Maybe the set of escaped characters could be shrinked and that would improve things.
  - some functions are not monomorphic, like numbers are being sent to be html-escaped too. It's difficult to optimize on this without any types at all, perhaps if there's any plan to add types to markdown-it there's something that could be done here.
  - the shape of tokens ("hidden class" as v8 calls it) changes too at runtime, I'm not sure how much there is to gain from never changing it, but it may be worth trying something in that regard.
  - overall I just don't think that there are some non-high-effort optimizations that would make the library twice as fast from what it is now.
- Far future
  - at the end of the day dealing with strings seems to be the bottleneck here, we just don't have enough control over them to make manipulating them faster. Perhaps the entire Markdown string should be transformed to bytes, those bytes/numbers should be manipulated directly, and then they could be transformed back to strings at the end of the process.
  - if the compiler worked mostly just with numbers, and didn't use regexes, perhaps it could be ported to WASM with some reasonable amount of effort.
  - if the compiler worked from WASM there might be something interesting that would be possible to do there with SIMD instructions, there seem to be at least a Markdown [compiler](https://github.com/raphlinus/pulldown-cmark) that uses them.

##### End

I'd probably just port over the modified "normalize" rule to get most of the performance benefits with ~none of the noise.

Sorry for the long write.

### DIFF
--- a/benchmark/benchloop.js
+++ b/benchmark/benchloop.js
@@ -1,0 +1,37 @@
+
+'use strict';
+
+/* IMPORT */
+
+var fs = require ('fs'),
+    path = require ('path'),
+    benchmark = require ('benchloop'),
+    mit = require ('..');
+
+/* HELPERS */
+
+var SAMPLES_PATH = path.join (__dirname, 'samples');
+var md = mit ('commonmark');
+
+/* BENCHMARK */
+
+benchmark.defaultOptions = Object.assign (benchmark.defaultOptions, {
+  iterations: 50,
+  log: 'compact'
+});
+
+fs.readdirSync (SAMPLES_PATH).forEach (function (sample) {
+
+  var samplePath = path.join (SAMPLES_PATH, sample),
+      content = fs.readFileSync (samplePath, 'utf-8');
+
+  benchmark ({
+    name: sample,
+    fn: function () {
+      md.render (content);
+    }
+  });
+
+});
+
+benchmark.summary ();
--- a/benchmark/profile.js
+++ b/benchmark/profile.js
@@ -14,6 +14,8 @@ var md = require('../')({
 // var data = fs.readFileSync(path.join(__dirname, '/samples/lorem1.txt'), 'utf8');
 var data = fs.readFileSync(path.join(__dirname, '../test/fixtures/commonmark/spec.txt'), 'utf8');
 
-for (var i = 0; i < 20; i++) {
+console.time('profile');
+for (var i = 0; i < 200; i++) {
   md.render(data);
 }
+console.timeEnd('profile');
--- a/lib/common/utils.js
+++ b/lib/common/utils.js
@@ -13,6 +13,14 @@ function has(object, key) {
   return _hasOwnProperty.call(object, key);
 }
 
+var _repeat = String.prototype.repeat || function (count) {
+  return new Array(count + 1).join (this);
+};
+
+function repeat(str, count) {
+  return _repeat.call (str, count);
+}
+
 // Merge objects
 //
 function assign(obj /*from1, from2, from3, ...*/) {
@@ -237,7 +245,7 @@ function isMdAsciiPunct(ch) {
   }
 }
 
-// Hepler to unify [reference labels].
+// Helper to unify [reference labels].
 //
 function normalizeReference(str) {
   // Trim and collapse whitespace
@@ -302,6 +310,7 @@ exports.lib.ucmicro         = require('uc.micro');
 exports.assign              = assign;
 exports.isString            = isString;
 exports.has                 = has;
+exports.repeat              = repeat;
 exports.unescapeMd          = unescapeMd;
 exports.unescapeAll         = unescapeAll;
 exports.isValidEntityCode   = isValidEntityCode;
--- a/lib/common/utils.js
+++ b/lib/common/utils.js
@@ -128,25 +128,38 @@ function unescapeAll(str) {
 
 ////////////////////////////////////////////////////////////////////////////////
 
-var HTML_ESCAPE_TEST_RE = /[&<>"]/;
-var HTML_ESCAPE_REPLACE_RE = /[&<>"]/g;
-var HTML_REPLACEMENTS = {
-  '&': '&amp;',
-  '<': '&lt;',
-  '>': '&gt;',
-  '"': '&quot;'
-};
+var escapeHtml = (function () {
+  if ( typeof XMLSerializer === 'function' ) {
+    var serializer = new XMLSerializer ();
+    var attr = document.createAttribute ( 'attr' );
+    var re = /[&<>"]/;
+    return function escapeHtml ( str ) {
+      if ( !re.test ( str ) ) return str;
+      attr.value = str;
+      return serializer.serializeToString ( attr );
+    }
+  } else {
+    var HTML_ESCAPE_TEST_RE = /[&<>"]/;
+    var HTML_ESCAPE_REPLACE_RE = /[&<>"]/g;
+    var HTML_REPLACEMENTS = {
+      '&': '&amp;',
+      '<': '&lt;',
+      '>': '&gt;',
+      '"': '&quot;'
+    };
 
-function replaceUnsafeChar(ch) {
-  return HTML_REPLACEMENTS[ch];
-}
+    function replaceUnsafeChar(ch) {
+      return HTML_REPLACEMENTS[ch];
+    }
 
-function escapeHtml(str) {
-  if (HTML_ESCAPE_TEST_RE.test(str)) {
-    return str.replace(HTML_ESCAPE_REPLACE_RE, replaceUnsafeChar);
+    return function escapeHtml(str) {
+      if (HTML_ESCAPE_TEST_RE.test(str)) {
+        return str.replace(HTML_ESCAPE_REPLACE_RE, replaceUnsafeChar);
+      }
+      return str;
+    }
   }
-  return str;
-}
+})();
 
 ////////////////////////////////////////////////////////////////////////////////
 
--- a/lib/renderer.js
+++ b/lib/renderer.js
@@ -44,9 +44,14 @@ default_rules.fence = function (tokens, idx, options, env, slf) {
       highlighted, i, arr, tmpAttrs, tmpToken;
 
   if (info) {
-    arr = info.split(/(\s+)/g);
-    langName = arr[0];
-    langAttrs = arr.slice(2).join('');
+    if (!/\s/.test (info)) {
+      langName = info;
+      langAttrs = '';
+    } else {
+      arr = info.split(/(\s+)/g);
+      langName = arr[0];
+      langAttrs = arr.slice(2).join('');
+    }
   }
 
   if (options.highlight) {
@@ -55,7 +60,7 @@ default_rules.fence = function (tokens, idx, options, env, slf) {
     highlighted = escapeHtml(token.content);
   }
 
-  if (highlighted.indexOf('<pre') === 0) {
+  if (highlighted.slice(0, 4) === '<pre') {
     return highlighted + '\n';
   }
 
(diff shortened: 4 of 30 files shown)


puzrin:
> I'd probably just port over the modified "normalize" rule to get most of the performance benefits with ~none of the noise.

Sounds reasonable.

- `\n\r` regexp worth to impove.
- `\0` - not sure. May be, `if \0 exists => call .replace`. Curent implementation cause sliced string store - potential roblem with speed for next scans.

fabiospampinato:
> \0 - not sure. May be, if \0 exists => call .replace. Curent implementation cause sliced string store - potential roblem with speed for next scans.

I think pretty much always `\0` won't be found in the string anyway, so further optimizations on this might not be that useful, plus `\0` in strings are a non-issue in JS so I'd remove the replacement entirely and squeeze some more performance out of the library. 

Removing the replacement entirely would only cause issues to people reading markdown-it's output from languages like C, and only if the input string contained a `\0` to being with basically, I don't think that happens in practice and those folks have always the option to perform the replacement themselves.

Just my opinion on this, maybe being fully compliant even with the weird parts of CM is worth more.

puzrin:
If \0 replace is from spec requirement, it can not be dropped. I don't remember details.

fabiospampinato:
Yeah the spec mentions somewhere that it must be replaced for "security reasons" IIRC, perhaps you could nudge the CM people to change the wording on that in v0.30 of the spec. There are really no security issues with it whatsoever in JS 🤷‍♂️

j2L4e:
> There are really no security issues with it whatsoever in JS

It depends on what you do with the output and how you display or store it, though. If `\0` sensitive services are involved, this might very well be an issue.

I think an option `replaceNullChars` with default `true` would do. It would also be a non-breaking change and could be documented properly (being not spec-compliant and potentially dangerous)

fabiospampinato:
> It depends on what you do with the output and how you display or store it, though. If \0 sensitive services are involved, this might very well be an issue.

By that logic if you think about it any string could be a security issue.

The reality is that this is a JS library, `\0`s aren't a problem at all in JS strings, and if other languages have some problems with those they should be the ones performing the sanitization on untrusted things. This isn't a C string sanitization library.

> I think an option replaceNullChars with default true would do. It would also be a non-breaking change and could be documented properly (being not spec-compliant and potentially dangerous)

That'd be too obscure and ugly IMO, plus there kind of is an option for this already since the "normalize" rule can be overridden entirely, but I would imagine approximately 0 people would even know that they can get a double-digit performance percentage boost by not replacing null characters.

hukkinj1:
> If \0 sensitive services are involved, this might very well be an issue.

> > By that logic if you think about it any string could be a security issue.

I think @j2L4e main argument wasn't really that there is a security issue. Just that some services may be sensitive to changes that are not CommonMark compliant. I myself have written code that relies on the assumption that Markdown parsed by markdown-it can never contain a null character, and I don't think I'm the only one.

Markdown-it advertises 100% CommonMark support in big fat letters so IMO any performance optimizations should be subordinate.

fabiospampinato:
> I myself have written code that relies on the assumption that Markdown parsed by markdown-it can never contain a null character, and I don't think I'm the only one.

Then you are doing it wrong IMO, for multiple reasons:

- CommonMark can change, if they word the silly requirement about the `\0` character slightly differently then markdown-it may stop replacing that, breaking whatever you are doing with it.
- Markdown-it is not infallible, every now and then there are bugs, how do you even know that the `\0` character is being replaced correctly every single time? Approximately 0 people would dig through the codebase to make sure of that.
- If you are passing the string compiled by markdown-it to a service that is sensitive to null characters then you should be treating that string as an unsafe input in the first place and perform the sanitization right before passing on the string, not doing that leaves you open to potential issues, a couple of which I mentioned in the other bullet points.

hukkinj1:
> Then you are doing it wrong IMO

If you say so, lol :smile: 

> CommonMark can change

Yup, it's very normal that in case the spec changes, then code does as well, to follow the updated spec. Should I build code that is prepared for any possible spec change? I'm not a clairvoyant, don't think that's possible.

> Markdown-it is not infallible, every now and then there are bugs

So I should build my code with the assumption that all downstream libraries are buggy. And not only that, but predict what bugs exactly will be introduced, and write code that is prepared beforehand?

> f you are passing the string compiled by markdown-it to a service that is sensitive to null characters...

I'm not.

fabiospampinato:
> Should I build code that is prepared for any possible spec change? I'm not a clairvoyant, don't think that's possible.

I'm only referring to the `\0` issue here, so yes, absolutely, if that's a problem for your use case you should be handling that outside of markdown-it's scope, regardless of what markdown-it does.

> So I should build my code with the assumption that all downstream libraries are buggy. And not only that, but predict what bugs exactly will be introduced, and write code that is prepared beforehand?

Again, regarding the `\0` issue, which is the only thing I was proposing markdown-it broke away from CM for, yes, you should be doing the replacement yourself.

> > f you are passing the string compiled by markdown-it to a service that is sensitive to null characters...
> I'm not.

Well I guess I misunderstood this sentence then:

> I myself have written code that relies on the assumption that Markdown parsed by markdown-it can never contain a null character, and I don't think I'm the only one.

It's good that what markdown-it does with `\0` characters is not a problem anymore for you, that's all I was suggesting, not relying on markdown-it for something like that.

hukkinj1:
> I'm only referring to the \0 issue here, so yes, absolutely, if that's a problem for your use case you should be handling that outside of markdown-it's scope, regardless of what markdown-it does.

So when markdown-it declares itself 100% CommonMark compliant, I should still make my own calls on what part of the spec its actually not? :smile: 

I'm not saying markdown-it can't stop removing null characters. I'm saying if it does, that should be a breaking major version, and it should be advertised 99% CommonMark compliant afterwards.

fabiospampinato:
> So when markdown-it declares itself 100% CommonMark compliant, I should still make my own calls on what part of the spec its actually not? 😄

When concerning security issues like the sanitization of null characters, yes, of course.

> I'm not saying markdown-it can't stop removing null characters. I'm saying if it does, that should be a breaking major version, and it should be advertised 99% CommonMark compliant afterwards.

It seems the only way markdown-it could make that change would be if CommonMark itself changed in the first place, in which case it will still be ~100% compliant.

j2L4e:
@fabiospampinato I thought about it some more as i think I kicked this off in the first place. I'm with you on that any string-parsing part of a system _should_ properly sanitize its own input and not rely on input to be safe to use.

Making this opt-in would be spec-compliant, safe(r) by default and non-breaking. Sorry for repeating it, I think it's the obvious way to go here.

```
const md = require('markdown-it')({
  nonCommonmark: {
    skipNullCharSanitization: true,
  }
});
```

fabiospampinato:
I've got a new freshly minted [optimization](https://github.com/markdown-it/markdown-it/pull/748/commits/bd779461d7ffad5a5848912d562d35954758c4ef) for you boys and girls.

Turns out there was a way to make the `escapeHtml` function, which is responsible for about ~30% of the time needed by markdown-it's benchmark, about ~65% faster, but it only works if you are in a browser environment.

In practice markdown-it's benchmark seems a little too heavy on escaping to me, if you also need to compile markdown files that require a lot of escaping in practice this optimization may speed up your use case by more than ~30%, but if 99% of the time your strings don't actually need to be html-escaped then this will be just as fast as before and not a bit faster.

@puzrin You might want to consider porting over this optimization too, as it's pretty significant and requires a fairly minimal change to the codebase.

puzrin:
@fabiospampinato i don't like idea of env0dependent code. That reduces maintainability.

Do you really has cases, where:

1. % of escaped strings is high
2. Total performance of parser is not enougth, and patch helps significantly

At first glance, it's a nice hack, but not convenient (difficulties are not compensated by added value).

fabiospampinato:
> That reduces maintainability.

I could move that code to a tiny dependency that I'll maintain if you like, in practice it will never really need to be changed, unless somebody finds a way to speed it up even more.

Overall whether this is worth it or not for "markdown-it" itself is your choice, I don't really care either way as I'll run a fork anyway and eventually I'll most probably, resources permitting, want to explore rewriting the all thing from scratch for better performance. I just wanted to share the patch as it might be useful to somebody else too.

For what is worth my use case is heavily depending on parsing and compiling Markdown, I could use every drop of performance I could squeeze out of this thing.

puzrin:
If that's just for info - no problem. As engineer, i like ALL you findings, those are interesting. But "as project manager" my opinion is sometime different. Because i have a lot of other projects and only 24 hours in a day :).

fabiospampinato:
Sure, I can totally understand and respect that 👍 

Let me maybe close this PR to keep the issue tracker clean.

I'll still be posting new findings in here if I'll happen to find any new ones.

fabiospampinato:
Kind of off-topic, but I'll mention this just in case:

Turns out that if you are running markdown-it in a browser you can make it ~23% smaller (according to bundlephobia), by replacing the "entities" dependency that is used for decoding HTML entities with something like the following, reusing the browser's own list of entities rather than shipping a list yourself:

```ts
const decode = entity => {
  const div = document.createElement ( 'div' );
  div.innerHTML = entity;
  return div.textContent;
};
```

puzrin:
I know principle of this hack. Prefer exactly the same behaviour in browser & node, until can't be avoided.

PR #762, opened 2021-03-08
Configurable typographer

kemitchell:
This pull request extends the `typographer` configuration option to allow users to provide an Object, rather than just`true`, with flags to toggle specific categories of replacements on and off. I have made rather granular commits, in case that aids review, but welcome a squash-merge.

This work comes out of discussions about the Discourse Internet discussion forum project, where we've found that some replacements, especially `(c)` &rarr; &copy;, trip up users, especially when copying lower-alpha enumerated lists:

```markdown
Common fruits include (a) apples, (b) bananas, (c) copyrights, and (d) dates.
```

Rendered:

> Common fruits include (a) apples, (b) bananas, &copy; copyrights, and (d) dates.

Desired:

> Common fruits include (a) apples, (b) bananas, (c) copyrights, and (d) dates.

These kinds of lists are especially common in outlines, statutes, contracts, and policies.

At the same time, other replacements, like typographers' quotes and dashes, work well for users, and don't cause frustrating problems quoting existing texts in Markdown.

I have no strong feelings about the format of the of Object option. I'm also open to revising this PR to allow a function, rather than a configuration option for the existing replacer configuration. I regret I'm not familiar enough with the stability guarantees around the parser APIs that would expose to know offhand whether that would be more complex or less so.

This PR is my original work, not within the scope of any IP arrangement with the company behind Discourse or any other. I'm happy to license it under the same MIT terms under which the module is currently licensed, waiving any right to have a copyright notice with my name included in copies and substantial portions.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -153,8 +153,8 @@ function normalizeLinkText(url) {
  * - __typographer__  - `false`. Set `true` to enable all [language-neutral
  *   replacement](https://github.com/markdown-it/markdown-it/blob/master/lib/rules_core/replacements.js) +
  *   quotes beautification (smartquotes). Or use an Object to enable
- *   specific replacements, like `{rare: true, ipSymbols:
- *   true, quotes: true, dashes: true}`.
+ *   specific replacements, like `{rare: true, ipSymbols: true, quotes: true,
+ *   dashes: true}`.
  * - __quotes__ - `“”‘’`, String or Array. Double + single quotes replacement
  *   pairs, when typographer enabled and smartquotes on. For example, you can
  *   use `'«»„“'` for Russian, `'„“‚‘'` for German, and
--- a/lib/index.js
+++ b/lib/index.js
@@ -154,7 +154,7 @@ function normalizeLinkText(url) {
  *   replacement](https://github.com/markdown-it/markdown-it/blob/master/lib/rules_core/replacements.js) +
  *   quotes beautification (smartquotes). Or use an Object to enable
  *   specific replacements, like `{rare: true, intellectualProperty:
- *   true, quotes: true}`.
+ *   true, quotes: true, dashes: true}`.
  * - __quotes__ - `“”‘’`, String or Array. Double + single quotes replacement
  *   pairs, when typographer enabled and smartquotes on. For example, you can
  *   use `'«»„“'` for Russian, `'„“‚‘'` for German, and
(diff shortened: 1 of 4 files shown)


kemitchell:
A side note: I did not attempt to implement CLI flags leveraging this new configurability. I thought it best to stop here and share first.

puzrin:
I strongly don't like idea to make options more complex. And i see no reasons why this can not be done as plugin and published on github

kemitchell:
Reimplementing as a separate module would mean copying or duplicating the existing replacement code implementation in a new plugin. For client-side applications using client-side bundlers, that would mean sending both across the wire, while disabling the implementation shipped in this module.

I had a quick look at the API documentation, but didn't see any notes about whether the APIs used by the existing replacement implementations fall under the stability guarantee. Maybe I looked in the wrong place?

puzrin:
https://github.com/markdown-it/markdown-it/blob/master/lib/rules_core/replacements.js replacement code is minimal. I don't like talks about code/speed optimizations without metrics. In this concrete case - size can not be reason for changes, IMO.

> I had a quick look at the API documentation, but didn't see any notes about whether the APIs used by the existing replacement implementations fall under the stability guarantee. Maybe I looked in the wrong place?

Please, understand me right, current api simplicity is result of rejecting new options. Because user's imagination is unlimited :). Default policy is "everything rejectable should be rejected" :). It's intended to direct everything to plugins.

When you suggest new option, you are expected to explain why this is very big exclusion. For example:

- required by 100% of users
- something in core is highly conflicting and blocks plugin implementation.

kemitchell:
I understand minimalism and stability. I practice the same in a number of modules I maintain, especially those used by the npm CLI. Though perhaps not to this extent.

It's your module, and I'm grateful for it. I don't want to bother you with arguments about changing things you don't want to change. I'd rather just express my thanks for your work here, and stop taking your time. :bow: 

My recommendation for Discourse will likely be to float the patch or ship a friendly fork removing all inbuilt linkify and typographer functionality, and reimplementing as plugins. That might be an opportunity to port up to ES Modules, as well.

puzrin:
As far as i remember, the most size-annoying things are linkifier and entities data.

- Linkifier can not be dropped (everybody wish it). And size reduce is possible only with new brosers, supporting unicode classes in regexp.
- Entities map could be significantly reduced, but commonmark spec requires full support.

So, ES Modules rewrite just will not change anything notable - loaded modules will stay the same.

PR #767, opened 2021-03-26
Fix table/list parsing ambiguity

hukkinj1:
Fix parsing of
```markdown
a | b
- | -
1 | 2
```

Markdown-it parses it as a table, but it should not be one, because there is a list on the second line that intercepts the table.

The GFM spec says:
> A leading and trailing pipe is also recommended for clarity of reading, and if there’s otherwise parsing ambiguity.

Also GitHub's own parser parses this as a list, not a table.

### DIFF
--- a/lib/rules_block/table.js
+++ b/lib/rules_block/table.js
@@ -52,7 +52,7 @@ function escapedSplit(str) {
 module.exports = function table(state, startLine, endLine, silent) {
   var ch, lineText, pos, i, l, nextLine, columns, columnCount, token,
       aligns, t, tableLines, tbodyLines, oldParentType, terminate,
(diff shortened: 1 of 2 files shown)


laurent22:
This actually introduces a regression for everybody who was writing their table without a leading pipes. See [here](https://github.com/laurent22/joplin/issues/7858) for example. Is this change really necessary? If there's a parsing ambiguity, then this ambiguity still exists after this change, and the only difference is that it broke a lot of existing Markdown documents.

rlidwka:
@laurent22 answered in https://github.com/laurent22/joplin/issues/7858#issuecomment-1563853366

PR #768, opened 2021-03-27
Switch default html option to false (safer) for CommonMark

zoafile:
Allowing html tags by default for the CommonMark preset is unsafe. Definitely got me surprised when I wrote an injection test to actually be vulnerable by default. Explicitly passing false works. This also lines it up with the README. 

Example below, I'm working downstream in the ruby port:

```
irb(main):003:0> MarkdownIt::Parser.new(:commonmark).render('<script>alert();</alert>')
=> "<script>alert();</alert>"
irb(main):004:0> MarkdownIt::Parser.new(:commonmark, { html: false }).render('<script>alert();</alert>')
=> "<p>&lt;script&gt;alert();&lt;/alert&gt;</p>\n"
```

### DIFF
--- a/lib/presets/commonmark.js
+++ b/lib/presets/commonmark.js
@@ -5,7 +5,7 @@
 
 module.exports = {
   options: {
-    html:         true,         // Enable HTML tags in source
+    html:         false,        // Enable HTML tags in source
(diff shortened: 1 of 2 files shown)


puzrin:
Commonmark preset exists for 100% spec compatibility. Personal opinions here are not acceptable.

PR #769, opened 2021-03-27
Allowlisting data:image/svg+xml

compulim:
Today, `image/gif`, `image/jpeg`, `image/png`, `image/webp` are allowed by [the default `validateLink`](https://github.com/markdown-it/markdown-it/blob/master/lib/index.js#L33).

As [all browsers (IE 9+) now support SVG](https://caniuse.com/?search=svg), to push vector graphics and hi-res display, we should consider enabling SVG by default.

This pull request is to add `data:image/svg+xml` to the allowlist, enabling SVG in images, such as: `Logo`. Also added tests to verify how the default link validator should work.

Mozilla has a great list of image types supported in browsers, https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img.

IANA official list of image types can be found at https://www.iana.org/assignments/media-types/media-types.xhtml#image. Note some modern formats such as [HEIF are not supported yet](https://caniuse.com/?search=heic).

Please let me know if you want to include more, such as [Animated PNG](https://caniuse.com/?search=apng), [AVIF](https://caniuse.com/?search=avif), BMP, ICO, and TIFF. Based on browsers support, I think APNG is next logical move to allowlist, but document author maybe unlikely to put it under data URI as the file could be huge.

Thanks.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -30,7 +30,7 @@ var config = {
 //
 
 var BAD_PROTO_RE = /^(vbscript|javascript|file|data):/;
-var GOOD_DATA_RE = /^data:image\/(gif|png|jpeg|webp);/;
+var GOOD_DATA_RE = /^data:image\/(gif|png|jpeg|svg\+xml|webp);/;
 
 function validateLink(url) {
   // url should be normalized at this point, and existing entities are decoded
--- a/test/misc.js
+++ b/test/misc.js
@@ -314,7 +314,9 @@ describe('Links validation', function () {
   it('default should skip blocklisted protocols', function () {
     var md = markdownit();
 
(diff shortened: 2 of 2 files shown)


puzrin:
Current limitations are not about features, but about security.

1. You proposal says nothing about security.
2. It's breaking change.
3. Validator can be replaced without mainstream change.

=> not enough reasons to merge. Also, worth to discuss breaking proposals first.

panta82:
For reference, here is how to opt into this behavior.

```javascript
const md = MarkdownIt({
  //...
});

md.validateLink = (originalValidateLink => url => {
    if (originalValidateLink(url)) {
      return true;
    }

    if (/^data:image\/svg\+xml;/.test(url)) {
      return true;
    }

    return false;
})(md.validateLink);
```

PR #783, opened 2021-05-20
fix: [rules:table] Missing `map` property

mokaddem:
Hello!
Thanks a lot to the team for building this project and maintaining it!

This PR fixes the following:
The `map` property of some tokens from the table rule are not set and thus are defaulted to `null`.


puzrin:
IMO duplicating rows info to each cell is overkill. Until column value (of [line, column]) is not supported, that looks useless.

mokaddem:
Hey,
In the end, I ended up looking for the row in which the cell is contained and relied on this.
I'm closing the PR then.
Thanks for getting back though.
Cheers!

PR #797, opened 2021-07-07
Parser: Set ordered list_item_open token info to input marker.

kohler:
Rendering plugins may be interested in the actual values of the markers in an ordered list. For instance, in a list

    1. Item 1
    20. Item 2

a plugin might want to render the input markers, 1 and 20, rather than the auto-incremented marker values (1 and 2).

This commit uses the `Token.info` slot to store the input markers for `list_item_open` tokens. In the example above, the first `list_item_open` token will have `token.info === '1'`, and the second will have `token.info === '20'`. The default renderer is unaffected, but plugins can access `token.info` if they are interested.

Benchmark performance seems similar.

### DIFF
--- a/lib/rules_block/list.js
+++ b/lib/rules_block/list.js
@@ -161,7 +161,7 @@ module.exports = function list(state, startLine, endLine, silent) {
   if ((posAfterMarker = skipOrderedListMarker(state, startLine)) >= 0) {
     isOrdered = true;
     start = state.bMarks[startLine] + state.tShift[startLine];
-    markerValue = Number(state.src.substr(start, posAfterMarker - start - 1));
+    markerValue = Number(state.src.slice(start, posAfterMarker - 1));
 
     // If we're starting a new ordered list right after
     // a paragraph, it should start with 1.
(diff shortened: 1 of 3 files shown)


rlidwka:
You've made `skipOrderedListMarker` return 2 values:
 - `pos` (normal return value)
 - `firstPos.pos` (via modifying object)

What is the purpose in that?

As far as I understand it, `firstPos.pos === pos - 1` always for successful parses, can't it be calculated from whereever the function is called?

rlidwka:
Not sure if `info` is appropriate (could do either `{ info: '20', markup: '.' }` or `{ info: '', markup: '20.' }`)... Nevermind, info is probably better for this.

kohler:
For an ordered list with more than one-digit markers, like “2000.”, `firstPos !== pos - 1`.

kohler:
I'd be happy to put this in `markup` but wasn't clear whether existing plugins depended on `markup === '.'`.

rlidwka:
> But yes you're right that using start, the way the code was written before, is equivalent, and I'll do that instead!

please do; changes look good to me otherwise

PR #799, opened 2021-07-12
Always suffix indented code block with a newline

hukkin:
Currently
```markdown
    codeblock\n
```
(where `\n` represents a trailing newline) renders differently to
```markdown
    codeblock
```
(no trailing newline).

The former renders to
```html
<pre><code>codeblock
</code></pre>
```
and the latter to
```html
<pre><code>codeblock</code></pre>
```

This PR makes the two examples render to the same HTML by always adding a trailing newline to code block content.

The spec is not overly clear about this corner case, but this PR makes
- the behavior in line with how the reference implementation parses and renders https://spec.commonmark.org/dingus/?text=%20%20%20%20codeblock
- makes it so that any indented code block can be reformatted as code fence (useful for markdown formatters)

### DIFF
--- a/lib/rules_block/code.js
+++ b/lib/rules_block/code.js
@@ -27,7 +27,7 @@ module.exports = function code(state, startLine, endLine/*, silent*/) {
   state.line = last;
 
   token         = state.push('code_block', 'code', 0);
-  token.content = state.getLines(startLine, last, 4 + state.blkIndent, true);
+  token.content = state.getLines(startLine, last, 4 + state.blkIndent, false) + '\n';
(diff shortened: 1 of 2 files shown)


rlidwka:
Merged, thanks!

Is code block the only place where the absense of trailing newline in original input matters?

hukkin:
> Is code block the only place where the absense of trailing newline in original input matters?

This I don't have an answer to right now, I haven't tested. I found this issue while fuzzing [mdformat](https://github.com/executablebooks/mdformat) and if there are other similar issues with other syntax then the fuzzer should find them when I have the time to get back into it.

PR #817, opened 2021-10-16
Typo fix

himanshumehta1114:


puzrin:
This doesn't look like typo fix only, because change regexp. Please split.

PR #829, opened 2021-12-12
Fix Simple example missing var  in README

slidoooor:


puzrin:
That example is correct

PR #853, opened 2022-03-21
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.

### DIFF
--- a/lib/rules_block/table.js
+++ b/lib/rules_block/table.js
@@ -9,7 +9,7 @@ function getLine(state, line) {
   var pos = state.bMarks[line] + state.tShift[line],
       max = state.eMarks[line];
 
-  return state.src.substr(pos, max - pos);
+  return state.src.slice(pos, max);
 }
 
 function escapedSplit(str) {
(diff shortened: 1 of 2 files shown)


jimmywarting (approved):

puzrin:
@CommanderRoot deprecation is usually not a big deal. For example `atob` & `btoa` are deprecated > 10 years, but still supported. Do you have real problem with `.substr()`?

puzrin:
Then i can stay this open to land later somehow, with other changes. Any such change can have side effects and requires benchmarking.

CommanderRoot:
From experience in other repos I can say that `slice` and `substr` are basically the same in regards to performance. Both are faster than `substring` however.

rlidwka:
merged, thanks

Issue #855, opened 2022-03-26
Doing some modernization?


jimmywarting:
Would you be open if i created some PR that refactor some things into es6 classes and using some const/let?

puzrin:
It's not a problem to refactor, but everything should have concrete reasons. Could you explain your ones?

jimmywarting:
Just want to learn the in and outs of this package of how it works and starting to contribute to new features eventually.

by using classes instead of prototyping vs code will be able to better understand and give better IDE auto suggestion as well...

puzrin:
Please, understand me right. I don't like to change development process without strong reasons. Things are unified across many repos, and i don't like to spend extra efforts for maintenance without big added value.

jimmywarting:
Ok... no worries.
just thought since you are testing node v12 that it would be fine...

https://github.com/markdown-it/markdown-it/blob/6da01033aa6ea2892e16a44672431fad3aff37b2/.github/workflows/ci.yml#L16

May I ask what targeted browser/node do you support?

if such things as this isn't alright, then i guess it isn't fine to even try to introduce classes, async/await or even function generators that could be part of a streaming approche to even try and introduce some form of `async highlight() { ... }` functionality...

puzrin:
All with es5 support should be ok.

PR #856, opened 2022-03-26
refactor: added the no-var eslint rule

jimmywarting:
## Purpose
To modernize the codebase a bit

## Changes
- Turned on the error in eslintrc.yml for the `no-var` rule
  - by doing so it started to complain about reserverd keyword, so i had to set es6 = true also
- executed `npx eslint . --fix` to auto fix the issues

---

- part of #855

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,7 +1,7 @@
 env:
   node:     true
   browser:  false
-  es6:      false
+  es6:      true
 
 ignorePatterns:
   - coverage/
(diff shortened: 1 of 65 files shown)


puzrin:
Sorry, i don't like to decline PRs, but this one does not seems to add value (i name such changes "just for fun"). Also, this drops compatibility with ancient browsers and requires major release.

I'd suggest to keep intact any things, if those could touch maintenance process. At least, without preliminary discuss in tracker.

PR #863, opened 2022-04-12
Support ES Module output in npm package. #862

### LINKED ISSUE #862: ES Module (ESM) build in the npm package
Like #852 but for the npm package version, as I do not use the jsDelivr build. (I have a transitive npm dependency on markdown-it via prosemirror-markdown.)

When trying to use this library in an Angular application, we get the following compiler warning:
```
Warning: [redacted_file_path] depends on 'markdown-it/lib'. CommonJS or AMD dependencies can cause optimization bailouts.
For more info see: https://angular.io/guide/build#configuring-commonjs-dependencies
```

It is now commonly recommended to support ES Module (ESM) builds in addition to legacy formats, which also has the added benefit of better bundle optimization. All supported versions of Node.js at the time of this writing fully support ES Modules, with the exception of Node 12 which has it behind an experimental flag, but that version is going out of support at the end of this month (April 2022).

The UMD/CommonJS build can be retained for those users that need it, but having an ESM build in addition to the UMD build would be ideal.

I'd be happy to take a stab at this, unless one of the core team members would prefer to do it. Thanks!

paulirwin:
Adds ES Module (ESM) output in the Rollup config (in addition to the existing UMD output), and adds a helpful `npm run build` script. Fixes #862

Per the `CONTRIBUTING.md` file, I am not including the ESM build here, but I can easily add that if you would like.

Example Angular component using the ESM build (add to package.json as a file-path dependency) which does not have the warning about optimization bailouts:

```typescript
import { Component, OnInit } from '@angular/core';
import MarkdownIt from 'markdown-it';

@Component({
  selector: 'app-root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.scss']
})
export class AppComponent implements OnInit {
  title = 'markdown-it-ng-demo';
  text: string = "";

  ngOnInit(): void {
    const md = new MarkdownIt();
    console.log(md);
    this.text = md.render("# Hi!");
    console.log(this.text);
  }
}
```

Template HTML:
```html
<div [innerHTML]="text"></div>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,17 @@
   "license": "MIT",
   "main": "index.js",
   "module": "dist/esm/markdown-it.mjs",
+  "type": "module",
+  "exports": {
+    ".": {
+      "import": "./dist/esm/markdown-it.mjs",
+      "require": "./index.js"
+    },
+    "./*": {
+      "require": "./*",
+      "import": "./*"
+    }
+  },
   "bin": {
     "markdown-it": "bin/markdown-it.js"
   },
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
   "license": "MIT",
   "main": "index.js",
   "module": "dist/esm/markdown-it.mjs",
-  "type": "module",
   "exports": {
     ".": {
       "import": "./dist/esm/markdown-it.mjs",
--- a/package.json
+++ b/package.json
@@ -12,10 +12,12 @@
   "repository": "markdown-it/markdown-it",
   "license": "MIT",
   "main": "index.js",
+  "module": "dist/esm/markdown-it.mjs",
   "bin": {
     "markdown-it": "bin/markdown-it.js"
   },
   "scripts": {
+    "build": "rollup -c support/rollup.config.js",
     "lint": "eslint .",
     "test": "npm run lint && nyc mocha && node support/specsplit.js",
     "coverage": "npm run test && nyc report --reporter html",
--- a/support/rollup.config.js
+++ b/support/rollup.config.js
@@ -5,6 +5,28 @@ import nodePolyfills from 'rollup-plugin-node-polyfills';
 import pkg from '../package.json';
 import { terser } from 'rollup-plugin-terser';
 
+const unminifiedPlugins = [
+  // Here terser is used only to force ascii output
+  terser({
+    mangle: false,
+    compress: false,
+    format: {
+      comments: 'all',
+      beautify: true,
+      ascii_only: true,
+      indent_level: 2
+    }
+  })
+];
+
+const minifiedPlugins = [
+  terser({
+    format: {
+      ascii_only: true,
+    }
+  })
+];
+
 export default {
   input: 'index.js',
   output: [
@@ -12,31 +34,25 @@ export default {
       file: 'dist/markdown-it.js',
       format: 'umd',
       name: 'markdownit',
-      plugins: [
-        // Here terser is used only to force ascii output
-        terser({
-          mangle: false,
-          compress: false,
-          format: {
-            comments: 'all',
-            beautify: true,
-            ascii_only: true,
-            indent_level: 2
-          }
-        })
-      ]
+      plugins: unminifiedPlugins,
     },
     {
       file: 'dist/markdown-it.min.js',
       format: 'umd',
       name: 'markdownit',
-      plugins: [
-        terser({
-          format: {
-            ascii_only: true,
-          }
-        })
-      ]
+      plugins: minifiedPlugins,
+    },
+    {
+      file: 'dist/esm/markdown-it.mjs',
+      format: 'es',
+      name: 'markdownit',
+      plugins: unminifiedPlugins,
+    },
+    {
+      file: 'dist/esm/markdown-it.min.mjs',
+      format: 'es',
+      name: 'markdownit',
+      plugins: minifiedPlugins,
     }
   ],
   plugins: [


thiagodp:
Hello @paulirwin,
I suggest using [microbundle](https://github.com/developit/microbundle) for that. It provides different bundle formats with a minimum effort.

paulirwin:
@thiagodp I appreciate the input, but this project already uses Rollup, so changing that is outside the scope of this PR.

thiagodp:
@paulirwin Ok. JFI, Microbundle is based on Rollup, and you could continue using the existing plug-ins.

tdekoning:
@paulirwin any update on this? Would be a great addition to the library

paulirwin:
@tdekoning I also would like to see this PR merged 😉

pauleustice:
I'd also really like to see this merged please 🙏

andymac4182:
Any update on this? It would help resolve some issues I am having.

paulirwin:
Tagging the two people listed in the markdown-it org: @puzrin @rlidwka 

It appears the community is longing to get this fix merged into the project. I'd be happy to address any feedback with it if there's anything preventing this from getting merged. Let me know how I can be of any assistance. Thanks!

vritant24:
Would very much like to see this PR merged. Happy to help take this through if help is needed.

JannieT:
As a workaround, I'm using this: https://github.com/esm-bundle/markdown-it

paulirwin:
I am generally a patient person when it comes to things like this, but this PR has been open for over a year and is fairly simple. It has been over two months since I pinged the owners of this repo, with no response. There have been no merged PRs since August of 2022 as far as I can tell. Seriously asking, is this project dead?

treeder:
Mark has ESM builds, FYI: https://www.npmjs.com/package/marked

or in browser: 

```html
import { marked } from "https://cdn.jsdelivr.net/npm/marked/lib/marked.esm.js"
```

hutianyu2006:
OMG I simply didn't do too much searching, THANK YOU A LOT!!!



---Original---
From: "Travis ***@***.***&gt;
Date: Fri, Jun 30, 2023 02:39 AM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [markdown-it/markdown-it] Support ES Module output in npmpackage. #862 (PR #863)




 
Mark has ESM builds, FYI: https://www.npmjs.com/package/marked
 
or in browser:
 import { marked } from "https://cdn.jsdelivr.net/npm/marked/lib/marked.esm.js"

paulirwin:
@treeder Unfortunately this is a transitive dependency for me so that isn't an option for all cases 😦

marijnh:
In 2023, we'll also want an `exports` field in package.json, something like ...

```json
  "exports": {
    "import": "./index.js",
    "require": "./dist/esm/markdown-it.mjs"
  },
```

That makes sure resolution that doesn't use the non-standard `"module"` field (such as node's own) can still find the ESM code.

andymac4182:
This would be great to have supported!

adrian-branescu:
@paulirwin @marijnh 
With the following change:
```
  "exports": {
    "require": "./index.js",
    "import": "./dist/esm/markdown-it.mjs"
  },
```
will be this PR merged eventually?

paulirwin:
Re: @marijnh (I am forever indebted to you for your excellent work on ProseMirror, btw. Thank you for that library!)

> In 2023, we'll also want an `exports` field in package.json, something like ...
> 
> ```json
>   "exports": {
>     "require": "./index.js",
>     "import": "./dist/esm/markdown-it.mjs"
>   },
> ```
> 
> That makes sure resolution that doesn't use the non-standard `"module"` field (such as node's own) can still find the ESM code.

This is good to know, thanks! I don't spend a lot of time in Node directly and only use this package via Angular. Should `"type": "module"` be added as well, or would that break "legacy" users? Do we still need to support CommonJS now that it's almost 2024 and Node v20 is LTS?

> 
> (Just verified that this PR still applies cleanly, but node will load the CJS code when you import it. With the above change to package.json, it loads the ESM code, and everything seems to work.)
> 
> **Further edit**: Actually, adding this will forbid importing modules directly from the package by path, which is maybe a style you intend to support, given that there are separate minified files. Which modules are users expected to want to load directly?

I don't have any particular intentions if this was addressed to me, other than ensuring that optimization in Angular builds works. I am a little out-of-the-loop of current Node package conventions and standards. I know at one point in the past, it was fairly common to include minified builds with your package like markdown-it has done prior to my PR. Is that no longer the case? If so, that would certainly simplify the Rollup config and build. I'm eager to learn from the community here on any ways I can improve this PR.

Re: @adrian-branescu 
> @paulirwin @marijnh With the following change:
> 
> ```json
> "exports": {
>     ".": {
>       "import": "./dist/esm/markdown-it.mjs",
>       "require": "./index.js"
>     },
>     "./*": {
>       "require": "./*",
>       "import": "./*"
>     }
> },
> ```

What does the second export with the `*` accomplish, and why is that needed here? Looking to learn as I am unfamiliar with that.

> will be this PR eventually merged?

I am encouraged to see recent activity in September of this year on this project. It would be great if this could get merged at some point. Obviously as the PR author I would love to see it merged, but alas I am not a contributor to this repo. It sounds like I do need to brush the dust off of it before it is merged, though, with the addition of `exports`.

marijnh:
> Should "type": "module" be added as well, or would that break "legacy" users?

That should be fine, and help some tools that still use the old convention.

> What does the second export with the * accomplish, and why is that needed here? Looking to learn as I am unfamiliar with that.

Once you have an `exports` field, _only_ the files mentioned in it can be loaded from the package. So, since this package has more than just a top-level entry point, adding these wildcards makes those sub-paths accessible for direct require.

I would hope updating this PR and prodding a bit more will lead to a new release.

paulirwin:
@marijnh @adrian-branescu Thank you! I have made those changes to this PR. Quick testing of a new Node package that references it by file path seems to work fine at least for constructing a MarkdownIt. If either of you can help validate that I did this correctly and sufficiently, that would be appreciated.

adrian-branescu:
@paulirwin @marijnh
I would add to the `rollup.config.js` file the [external](https://rollupjs.org/configuration-options/#external) field, e.g.
```js
external: [
  /node_modules/
]
```
in order to prevent including in the bundle the external dependencies from package.json:
```json
"dependencies": {
    "argparse": "^2.0.1",
    "entities": "~3.0.1",
    "linkify-it": "^3.0.1",
    "mdurl": "^1.0.1",
    "uc.micro": "^1.0.5"
  },
```

In this way, one can use this NPM package in Node.js and Web apps with a bundler like webpack, Rollup, vite etc without having those dependencies included multiple times in the final app bundle.

marijnh:
>>   Should "type": "module" be added as well, or would that break "legacy" users?

> That should be fine, and help some tools that still use the old convention.

Correction! I thought you were talking about the `"module": "dist/esm/markdown-it.mjs"` line. `"type": "module"` should definitely **NOT** be added here.

paulirwin:
> @paulirwin @marijnh I would add to the `rollup.config.js` file the [external](https://rollupjs.org/configuration-options/#external) field, e.g.
> 
> ```js
> external: [
>   /node_modules/
> ]
> ```
> 
> in order to prevent including in the bundle the external dependencies (along with the whole graph of transitive dependencies) from package.json:
> 
> ```json
> "dependencies": {
>     "argparse": "^2.0.1",
>     "entities": "~3.0.1",
>     "linkify-it": "^3.0.1",
>     "mdurl": "^1.0.1",
>     "uc.micro": "^1.0.5"
> },
> ```
> 
> In this way, one can use this NPM package in Node.js and Web apps with a bundler like Webpack, Rollup, Vite etc without having those dependencies included multiple times in the final app bundle.

If I'm understanding this correctly, this would be an existing problem with the current rollup CJS output, is that correct? If so it seems like that warrants its own PR and could be merged independently.

marijnh:
Thanks. With that change this patch works for me (I tried: importing as ESM, requiring as CommonJS, requiring a specific dist/ file.)

adrian-branescu:
> > @paulirwin @marijnh I would add to the `rollup.config.js` file the [external](https://rollupjs.org/configuration-options/#external) field, e.g.
> > ```js
> > external: [
> >   /node_modules/
> > ]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in order to prevent including in the bundle the external dependencies (along with the whole graph of transitive dependencies) from package.json:
> > ```json
> > "dependencies": {
> >     "argparse": "^2.0.1",
> >     "entities": "~3.0.1",
> >     "linkify-it": "^3.0.1",
> >     "mdurl": "^1.0.1",
> >     "uc.micro": "^1.0.5"
> > },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In this way, one can use this NPM package in Node.js and Web apps with a bundler like Webpack, Rollup, Vite etc without having those dependencies included multiple times in the final app bundle.
> 
> If I'm understanding this correctly, this would be an existing problem with the current rollup CJS output, is that correct? If so it seems like that warrants its own PR and could be merged independently.

Yes, this problem replicates on the master branch too.

On your branch, there are >2000 lines of external source code added besides package own source code:
```bash

# find lib/ -name '*.js' | xargs wc -l
6185 total
 
# cat dist/markdown-it.js | wc -l
8356
 
# cat dist/esm/markdown-it.mjs | wc -l
8675
```

These may be duplicated in a Web App that makes use of the same dependencies like this package because the app bundler has no idea that they were already included in this package bundle.

paulirwin:
@adrian-branescu Thanks for the info, I could see where the addition of this could exacerbate that problem. If there's consensus from others that the `external` bit should be included here, I can do that.

marijnh:
@puzrin Are you okay with releasing this? Or giving some feedback on what would be needed before it gets released?

puzrin:
We closed most of pending things, and I'm rewriting to es6 to solve the rest.

Need advice, what is preferable:

- use `.mjs` extensions
- use `.js` with `type: module`

Could you help?

marijnh:
It doesn't matter much, but if you want to keep supporting the old behavior of requiring `dist/markdown-it.min.js`, you'll need to either use `.mjs` for your ES module files, or explicitly list all allowed paths in your package.json `exports` field, because with `type: module` a `.min.js` file would be interpreted as a module.

puzrin:
@marijnh thank you very much for info! I tend use `.msj`, if that will be ok to all next years.

puzrin:
See `es6` branch https://github.com/markdown-it/markdown-it/tree/es6

- all rewritten all to es6 (dependencies still not, but will be done)
- `dist/` now has umd build with all deps bundled.

Since change is huge, and I have no solid experience with esm development, I'm open to any suggestions, what should be improved. Hope to spread the same changes to all other maintained repos.

puzrin:
- Updated a lot of code to be more close to `standard` lint rules (var => let/const and so on).
- Added babel to es5 build, for sure.

@SReject The problem with classes is, that existing implementation uses hack for the call without `new`. This is not possible with `class`. I could export factory function as `default` and class as `MarkdownIt`. But not sure how good is that for users. Any suggestions welcome.

Also not sure, if should I use rollup+babel for the es6 modules build, or leave that for upper-level use.

puzrin:
I'm closing this, since master rewritten to ESM, and have fallback to CJS (generated on publish, or via `npm run build`).

It will take some time (up to 1-2 weeks) to check other issues and transform plugins. But you can write any thoughts if something else should be fixed.

puzrin:
Published v14 with native ESM support.

PR #864, opened 2022-04-18
fix(rules/fence): trim info trailling whitespaces

moki:
trim trailling whitespaces from info field of the fence token

### DIFF
--- a/lib/rules_block/fence.js
+++ b/lib/rules_block/fence.js
@@ -89,7 +89,7 @@ module.exports = function fence(state, startLine, endLine, silent) {
   state.line = nextLine + (haveEndMarker ? 1 : 0);
 
   token         = state.push('fence', 'code', 0);
-  token.info    = params;
+  token.info    = params.trim();
   token.content = state.getLines(startLine + 1, nextLine, len, true);
   token.markup  = markup;
   token.map     = [ startLine, state.line ];


rlidwka:
Can you please clarify why is it needed?

rlidwka:
Can you please clarify why is it needed?

moki:
leaving as it is, this code:

```
'''js____
console.log('hello, world');
'''
```
where

**'** means **`**
and
**_** means **whitespace**


would be parsed as:
```
Token {
    type: 'fence',
    info: 'js    '
}
```

parsing trailing whitespaces as part of the info field.

whatever is the pipeline you put parser in, one would get unnecessary data inside info field.

As an example i have a pipeline, where i feed `info`(which is programming language name) into library to make a decision
about how to render code, and that leads to an error `unknown language js____`.

where **_** means **whitespace**

rlidwka:
merged in https://github.com/markdown-it/markdown-it/commit/67fa0f239752ec649a3e93fe8bd7a3f410fec0f3 (with added changelog)

rlidwka:
nvm, reverted for now, not sure if breaking change is worth it

puzrin:
This change is breaking, and we should force trailing spaces to drop everywhere. The concrete example can be solved via renderer, without touching everything else.

Issue #865, opened 2022-04-20
How to enable xhtmlOut from the console?


ngcdev:
Hello!

First of all, thank you for this great tool! I'm using it to publish MD to HTML that is later uploaded to Confluence.

However, Confluence requires XHTML in their API. For the most part it's fine, but from time to time a file will fail. I examined the code and I noticed that you do have an option (internally) to output XHMTL (xhtmlOut):
```
  md = require('..')({
    html: !options.no_html,
    xhtmlOut: false,
    typographer: options.typographer,
    linkify: options.linkify
  });
```
Since I'm running it from the console ( `$ markdown-it SOME.MD > output.html`) I have not been able to figure out how to pass that parameter. I see that other options are taken from the "options" parameter, which are parsed from the console arguments, but there doesn't seem to be an option for the xhtmlOut.

An I missing something? How can I pass a parameter to output xhtml? I read that the require('..') can be used to load a file, but not sure if that will work here (to create my own js options file) (Sorry, not a node developer)

Thanks for your attention!

rlidwka:
`markdown-it` binary is created mostly for testing purposes and doesn't cover all possible options (presets, plugins, highlighter, etc.)

I would suggest to write custom wrapper and enable exactly what you need.

Something like this would do:

```js
#!/usr/bin/env node

const fs = require('fs')

let md = require('markdown-it')({
  html: false,
  xhtmlOut: true,
  typographer: true,
  linkify: true,
  breaks: false,
  langPrefix: 'lang-',
  // etc.
})

let source = fs.readFileSync('input.txt')
let output = md.render(source.toString())
fs.writeFileSync('output.txt', output)
```

Proposals to add all possible features (something like babel-cli does) are welcome.

PR #869, opened 2022-05-02
Adding argument to enable XHTML output

ngcdev:
Adding a new argument (`-x` or `--xhtml`) to enable XHTML output when calling markdown-it from the console.
Example:
```
$ markdown-it README.md -x >output.xhtml
```
This in en reference to issue/question https://github.com/markdown-it/markdown-it/issues/865

### DIFF
--- a/bin/markdown-it.js
+++ b/bin/markdown-it.js
@@ -30,6 +30,11 @@ cli.add_argument('-l', '--linkify', {
   action: 'store_true'
 });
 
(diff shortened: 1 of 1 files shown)


puzrin:
Closing as "rejected". We can consider design with "all possible options at once", but adding options one-by-one can go to unpredictable direction, and we avoid such approach, sorry. I'd suggest to just create your own custom wrapper, that's a fastest way.

PR #913, opened 2023-01-25
Moved `hljs` class to `<code>` from `<pre>`.

BMorearty:
In this change set I am changing the guidance in the README and in comments to put the class on the correct tag: `<code>`, not `<pre>`.

If you put the `hljs` class on `<pre>`, syntax highlighting does not use the correct padding.

You can see from the default CSS stylesheet in the highlight.js repo that the `hljs` tag is expected to be on the `<code>` tag:
https://github.com/highlightjs/highlight.js/blob/84719c17a51d7bb045f2df441b9c00f871f7c063/src/styles/default.css#L17-L25

All the official highlight.js examples have the `hljs` class on `<code>`:
https://highlightjs.org/static/demo/

This change doesn't affect the demo because (1) you overrode the `hljs` padding in your CSS and (2) even if you hadn't, you're using Bootstrap which also overrides `<pre>` padding.

Before change:
image

After change:
image

But here is an example from my own project where I discovered this issue. This project does not use Bootstrap or override the padding for `hljs`:

Before:
image

After:
image

I should also mention a couple of other issues with the README that I did not address:

1. Under "Syntax highlighting," it gives a default recommendation and then a second one: "Or with full wrapper override." But for syntax highlighting to use the correct background color with highlight.js, the second code sample is the only one that will work because the first code sample doesn't include the `hljs` class anywhere.
2. ~The last line of the second code sample looks like it has a bug. It references `md` while it's still in the value being assigned to `md`.~

I didn't try to address either of these issues but I can make a separate PR for them if you agree.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ var md = require('markdown-it')({
 });
 ```
 
-Or with full wrapper override (if you need assign class to `<pre>`):
+Or with full wrapper override (if you need assign class to `<pre>` or `<code>`):
 
 ```js
 var hljs = require('highlight.js'); // https://highlightjs.org
@@ -171,13 +171,13 @@ var md = require('markdown-it')({
   highlight: function (str, lang) {
     if (lang && hljs.getLanguage(lang)) {
       try {
-        return '<pre class="hljs"><code>' +
+        return '<pre><code class="hljs">' +
                hljs.highlight(str, { language: lang, ignoreIllegals: true }).value +
                '</code></pre>';
       } catch (__) {}
     }
 
-    return '<pre class="hljs"><code>' + md.utils.escapeHtml(str) + '</code></pre>';
+    return '<pre><code class="hljs">' + md.utils.escapeHtml(str) + '</code></pre>';
   }
 });
 ```
--- a/lib/index.js
+++ b/lib/index.js
@@ -197,7 +197,7 @@ function normalizeLinkText(url) {
  * });
  * ```
  *
- * Or with full wrapper override (if you need assign class to `<pre>`):
+ * Or with full wrapper override (if you need assign class to `<pre>` or `<code>`):
  *
  * ```javascript
  * var hljs = require('highlight.js') // https://highlightjs.org/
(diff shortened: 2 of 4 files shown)


laurent22:
I don't think styling should be a consideration when generating the HTML since clients will have their own stylesheet anyway? And making it look nice by default in particular browser (I assume it's Chrome here) shouldn't be an objective of the library.

More importantly, I feel that such a change could break many clients which may expect the tags in a certain order. I'm not saying this change is necessarily bad but we should take into consideration that the lib is used by thousands of projects, and that should be taken into account when considering a change

BMorearty:
My counterarguments to @laurent22's points:

This is a documentation fix. The changes consist of fixing the readme, comments, and sample usage code in order to help new users. I did not modify the default HTML generation. I fixed the documentation to be consistent with the CSS that the project already provides. I don't see how this would be considered a change in how styling is applied when generating HTML. And it's not a breaking change to fix the documentation and make it consistent with the library. This won't affect anyone who's already using the library—they can continue to use their own HTML generation and their own CSS if they want—but will help new adopters come up to speed faster with the provided default CSS, without trying to hunt down the source of styling bugs as I did.

I don't know why an assumption was made that I'm doing this to appease a particular browser. I discovered the bug in Firefox after following the docs to the letter, but it reproduced in Chrome and Safari.

IlCallo:
I agree with @BMorearty, this was helpful for me and will be for new users, it won't touch existing projects

IlCallo:
> Under "Syntax highlighting," it gives a default recommendation and then a second one: "Or with full wrapper override." But for syntax highlighting to use the correct background color with highlight.js, the second code sample is the only one that will work because the first code sample doesn't include the hljs class anywhere.

I guess that's because it shows you how to apply syntax highlight in general, you could use something else or a custom function. It would probably be better to change the first example to use some custom code instead of hljs

> The last line of the second code sample looks like it has a bug. It references md while it's still in the value being assigned to md.

That's valid JS, you can do it because `highlight` will actually access the `md` variable after it has been assigned, it's a pretty common pattern, even if it has some pitfalls

BMorearty:
> That's valid JS ...

Thanks for the correction. I've stricken this point.

PR #937, opened 2023-06-07
Remove spurious state attribute

chrisjsewell:
As far as I can see, this attribute is no longer used?

### DIFF
--- a/test/pathological.json
+++ b/test/pathological.json
@@ -1,1 +1,1 @@
-{ "md5": "da7a30ea2886aea813e50c2764ecc40f" }
+{ "md5": "68c7c28fe0d4a34367f27dcbceec9fedda7a30ea2886aea813e50c2764ecc40f" }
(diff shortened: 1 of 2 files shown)


test/pathological.json
```diff
@@ -1 +1 @@
-{ "md5": "da7a30ea2886aea813e50c2764ecc40f" }
+{ "md5": "68c7c28fe0d4a34367f27dcbceec9fedda7a30ea2886aea813e50c2764ecc40f" }
```
chrisjsewell, line 1:
```suggestion
{ "md5": "da7a30ea2886aea813e50c2764ecc40f" }
```

chrisjsewell:
I'm not sure why the pathological test is failing or how to fix it?

rlidwka:
> I'm not sure why the pathological test is failing or how to fix it?

The test checks that pathological tests in commonmark.js repo didn't change since last time a human synced them.

The way to fix it is to change the hash (and also hopefully verify that nothing important changed over there). Or propose to remove after arguing against its usefulness.

PR #991, opened 2024-01-10
Add svg+xml as valid data url format

mfix-stripe:
Based on feedback we received here: https://github.com/markdoc/markdoc/issues/479

### DIFF
--- a/lib/index.mjs
+++ b/lib/index.mjs
@@ -29,7 +29,7 @@ const config = {
 //
 
 const BAD_PROTO_RE = /^(vbscript|javascript|file|data):/
-const GOOD_DATA_RE = /^data:image\/(gif|png|jpeg|webp);/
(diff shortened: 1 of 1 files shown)


dpb587:
I added the following comment to the referenced issue that I initially started. I don't think svg+xml should necessarily be added as a default because I think there may be security implications to the role that the `validateLink` function is trying to play. FWIW, I'm not an expert in markdown-it or SVG.

> I don't think it's a good idea to add it as a default in markdown-it. Support of SVG does require mitigations for potential security vulnerabilities (like XSS). In the particular case of `` it shouldn't be a concern with browsers disallowing an execution context there. But since `validateLink` is a general-purpose function, I'd be wary that adding it would open up other SVG/XSS injection points, such as through regular links or, more likely, third-party plugins which may be relying on the function.

puzrin:
Please, do this on user side or create a plugin.

Issue #996, opened 2024-01-26
Major performance issue when parsing a long list of reference links


RomanHotsiy:
We noticed a major performance problem when parsing a long list of references similar to this benchmark: https://github.com/markdown-it/markdown-it/blob/master/benchmark/samples/block-ref-list.md

In our case we have a list of 1000+ references.

The root cause seems to be this termination logic:
https://github.com/markdown-it/markdown-it/blob/d07d585b6b15aaee2bc8f7a54b994526dad4dbc5/lib/rules_block/reference.mjs#L29-L51

Removing this logic doesn't break any tests and improves speed of parsing our long list 30x 🙀

I tried to find some similar problems and found this thread: https://github.com/markdown-it/markdown-it/issues/54

I believe this table is incorrect but I'm not sure:

https://github.com/markdown-it/markdown-it/blob/d72c68b520cedacae7878caa92bf7fe32e3e0e6f/lib/parser_block.js#L13-L25

From the CommonMark spec I can't see that reference can be terminated by other rules and it's the other way around actually - the reference can terminate some of the rules. Am I correct?

I tried modifying the code above to the variant below and all the tests are passing performance is still fast:

```js
const _rules = [
  // First 2 params - rule name & source. Secondary array - list of rules,
  // which can be terminated by this one.
  ['table',      r_table,      ['paragraph']],
  ['code',       r_code],
  ['fence',      r_fence,      ['paragraph', 'blockquote', 'list']],
  ['blockquote', r_blockquote, ['paragraph', 'blockquote', 'list']],
  ['hr',         r_hr,         ['paragraph', 'blockquote', 'list']],
  ['list',       r_list,       ['paragraph', 'blockquote']],
  ['reference',  r_reference, ['table', 'fence', 'blockquote', 'hr', 'list', 'html_block', 'heading']],
  ['html_block', r_html_block, ['paragraph', 'blockquote']],
  ['heading',    r_heading,    ['paragraph', 'blockquote']],
  ['lheading',   r_lheading],
  ['paragraph',  r_paragraph]
]
```

Could someone check if my understanding is correct? I would be happy to open a PR.

RomanHotsiy:
Now I read a bit more code and I'm not sure I understand the spec correctly 🤔

https://github.com/markdown-it/markdown-it/blob/d07d585b6b15aaee2bc8f7a54b994526dad4dbc5/lib/rules_block/reference.mjs#L166

RomanHotsiy:
[...]

PR #1004, opened 2024-02-04
fix quadratic complexity in reference parser

rlidwka:
should fix https://github.com/markdown-it/markdown-it/issues/996

instead of calling `state.getLines` on the entire thing (until empty line), it now pulls in lines as needed

@RomanHotsiy please check if there's any improvement on your data

### DIFF
--- a/lib/helpers/parse_link_destination.mjs
+++ b/lib/helpers/parse_link_destination.mjs
@@ -10,7 +10,6 @@ export default function parseLinkDestination (str, start, max) {
   const result = {
     ok: false,
     pos: 0,
-    lines: 0,
     str: ''
   }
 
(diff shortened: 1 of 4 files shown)


RomanHotsiy (approved):
Thanks @rlidwka!!! Sorry for the delay.

This is amazing idea! 
It improves the performance a lot! On my big list of reference links (1000 lines) it gets ~27x speed improvement 🔥 

**version from main**

```
Sample: block-ref-list.md (63653 bytes)
 > current x 16.89 ops/sec ±0.50% (45 runs sampled)
```

**version from this branch**

```
Sample: block-ref-list.md (63653 bytes)
 > current x 437 ops/sec ±0.27% (91 runs sampled)
```

PR #1009, opened 2024-02-24
Update to spec 0.31.2

notriddle:
The spec update changes these things:

* It simplifies the HTML regex so that `<!-- a -- b -->` is an HTML comment. HTML5 reports this as an error, but still parses it.
* It changes the set of known HTML block elements to match HTML5, adding `search` and removing `source`.
* It adds Unicode Symbols to the set of punctuation characters that are used to evaluate flankingness. **This is blocked on** https://github.com/markdown-it/uc.micro/pull/15

This commit also changes the declaration HTML regex to match lowercase, even though that change was technically made in spec version 0.30.


rlidwka:
> This is blocked on https://github.com/markdown-it/uc.micro/pull/15

Blocker should be resolved.

Can you please rebase and update pull request?

rlidwka:
Also, can you move actual code changes and spec update to separate commits, if that's too not much of a trouble? (i.e. one commit with `test/fixtures/commonmark/*` and another with everything else)

PR #1041, opened 2024-08-04
Change font

rivercory:
Hello!
I changed font to Poppins to improve the readability of the markdown-it document.
If you don't like my PR, please leave a review!


puzrin:
Sorry to decline. Those files are auto-generated, and should not be edited manually. Also, standard font stacks would be preferable to external dependency.

Issue #1050, opened 2024-09-13
`parse` failing to tokenize "[a]: a"


celestine-stripe:
Hello! I'm using `markdown-it` version 14.1.0 and ran into a few edge cases where `parse` is failing to produce any tokens. Here's a script to reproduce the bug:

```ts
import markdownit from 'markdown-it';

const md = markdownit();

const cases = [
  '[aaa]: aa aaa.',
  '[aaa]: aa aaa',
  '[aaa]:',
  '[aaa]: ',
  'aaa: aa',
  '[aaa: aa',
  '[aaa] a.',
  '[]:',
  '[]: \n',
  '[]: .',
  '[[aaa]: a',
  '[]: a',
  '[ ]: a',
  'a [a]: a',
  // Cases below fail
  '[aaa]: aa',
  '[aaa]: a.',
  '[a]: a',
  '[.]: a',
];

cases.forEach((text) => {
  if (md.parse(text, {}).length === 0) {
    console.log(`Failed to sucessfully parse "${text}"`);
  }
});
```

The output is:
```
Failed to sucessfully parse "[aaa]: aa"
Failed to sucessfully parse "[aaa]: a."
Failed to sucessfully parse "[a]: a"
Failed to sucessfully parse "[.]: a"
```

I can repro the behavior using https://markdown-it.github.io/ as well!

rlidwka:
What tokens did you expect to get?

----

It's a [link reference definition](https://spec.commonmark.org/0.31.2/#link-reference-definition). It's not supposed to produce any tokens itself, instead it affects how other links are parsed (no links = no output).

You can see results in env instead:

```js
import markdownit from 'npm:markdown-it';

const md = markdownit();

let env = {};

console.log(md.parse("[a]: a", env));
// []
console.log(env);
// { references: { A: { title: "", href: "a" } } }
```

I think there was a discussion to make it a token anyway (for roundtripping and such), but that's not a bug, and rather lack of feature.

celestine-stripe:
Oh interesting! Hm, so I was expecting:
```ts
md.parse('[aaa]: aa aaa', {})
```
to produce:
```ts
[
  {
    "type": "paragraph_open",
    "tag": "p",
    "attrs": null,
    "map": [
      0,
      1
    ],
    "nesting": 1,
    "level": 0,
    "children": null,
    "content": "",
    "markup": "",
    "info": "",
    "meta": null,
    "block": true,
    "hidden": false
  },
  {
[...]

PR #1055, opened 2024-09-30
Added link reference 'definition' token type

h7kanna:
Added a new token type for link reference definition.

I needed definitions in the Markdown AST(token stream). Based on this [comment](https://github.com/markdown-it/markdown-it/issues/1050#issuecomment-2365359497) and its inclusion in the [Rust port](https://github.com/markdown-it-rust/markdown-it/blob/master/src/plugins/cmark/block/reference.rs#L365) here, figured it will be acceptable to add this as a feature.

### DIFF
--- a/lib/renderer.mjs
+++ b/lib/renderer.mjs
@@ -187,7 +187,7 @@ Renderer.prototype.renderToken = function renderToken (tokens, idx, options) {
   //  - a
   //    >
   //
-  if (token.block && token.nesting !== -1 && idx && tokens[idx - 1].hidden) {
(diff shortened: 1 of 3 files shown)


puzrin:
That's potentially doable. But I'm not ready to accept that immediately, since the change is breaking and and i can't predict side-effects (for example, how to sync changes).

This is not a complete rejection. But i think we should discuss this first in issue, and plan for next major version.

Iwagd-DHA (approved):

puzrin:
Solved in `dev` branch (upcoming v15)

PR #1074, opened 2024-12-15
Don't trim non-ASCII whitespaces

### LINKED ISSUE #1067: Don't trim non-ASCII whitespaces
- https://github.com/commonmark/commonmark.js/issues/261
- https://github.com/commonmark/commonmark.js/pull/289

CommonMark doesn't require implementations to remove non-ASCII white spaces (just says "space", which means U+0020).
However, markdown-it removes them.
This is due to the fact markdown-it and prior CommonMark.js uses `.tirm()` to remove leading/trailing space-like characters, including non-ASCII ones.

The specification of `.trim()`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim

tats-u:
Fixes #1067


tats-u:
If you will not review this PR, I will publish a new markdown-it plugin package only to deal with this issue.

puzrin:
I think `asciiTrim` name will be more comprehensive. Ok to merge. Please add clarifications in https://github.com/markdown-it/markdown-it/issues/1067

puzrin:
Note. The same issue as in another PR. Function (isSpaceLike) should NOT be placed into another one for performance reasons.

Possible name ~ `isAsciiTrimmable`. After one month, everyone will forget about this PR, what `isSpaceLike` means, and why it exists at all. It's good when anyone can understand code immediately, without thinking. Probably, an additional comment will help.

`Light trim for blocks like paragraph/header, where simple whitespaces should be trimmed, but unicode ones should stay intact.`

tats-u:
Renamed based on your suggestion.

PR #1075, opened 2024-12-15
Use cross-env to run test

tats-u:
Windows can't run test because its shell (CMD.EXE) doesn't support the format `ENV_NAME=value command to be run`.
[cross-env](https://github.com/kentcdodds/cross-env) is a standard solution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   },
   "scripts": {
     "lint": "eslint .",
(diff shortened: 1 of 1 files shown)


kimsoo527 (approved):

tats-u:
Thank you for the merge. By the way, if you make it mandatory to use pnpm, you wouldn't need to depend on cross-env anymore.

https://pnpm.io/cli/run#shellemulator

Issue #499, opened 2018-10-10
access env object in highlight options


naivefun:
Let's say I need process fence code with async API, the suggested way is to use env to store temporary hint, but in highlight option there is no access to env, how can I do that please?

puzrin:
You can replace default fenced block rule to anything you wish.

PR #1119, opened 2025-08-03
feat: added `env` argument to `highlight` option

### LINKED ISSUE #1105: highlight 函数请求返回env对象
highlight 函数请求返回env对象，从而实现单例模式，后续只需要通过render函数渲染，传入不同的env就可实现多环境

serkodev:
Related #626
Fixed #1105, #499

Since the default `fence` rule callback has a `env` param, I think it would be better if it can pass to the `options.highlight` as well to provide additional context info.

### DIFF
--- a/lib/renderer.mjs
+++ b/lib/renderer.mjs
@@ -40,7 +40,7 @@ default_rules.fence = function (tokens, idx, options, env, slf) {
 
   let highlighted
   if (options.highlight) {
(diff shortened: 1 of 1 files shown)


puzrin:
I'm not comfortable with this API change. As has been said in the referred issues, modify the default fenced rules if you do not agree.

PR #1138, opened 2025-11-30
Refactor code highlighting function to use template literals

ShaiGerbi:


puzrin:
- This change adds unexpected linebreaks.
- There are many other places in such an old style. Imho, it isn't worth making single microchanges. Also, any style changes will be discussed first.

PR #1140, opened 2025-12-01
feat!: remove ddIndent property from state_block

### LINKED ISSUE #1139: outdated ddIndent in stateBlock
The ddIndent was introduced by https://github.com/markdown-it/markdown-it/commit/3f16868edd1372f65b30e7b821e55158204710fc, while it was lately removed and the  original file becomes a standalone plugin:https://github.com/markdown-it/markdown-it-deflist/blob/master/index.mjs

However, it makes no sense that the ddIndent should be preserved in StateBlock. And it's also weird that I cannot find any code that is changing the ddIndent state in markdown-it, only codes from markdown-it-deflist

I suggest removing it from StateBlock to avoid confusion. This property should be added and maintained by `markdown-it-deflist` internally in its own code

Mister-Hope:
closes https://github.com/markdown-it/markdown-it/issues/1139

### DIFF
--- a/lib/rules_block/state_block.mjs
+++ b/lib/rules_block/state_block.mjs
@@ -42,7 +42,6 @@ function StateBlock (src, md, env, tokens) {
   this.line       = 0 // line index in src
   this.lineMax    = 0 // lines count
   this.tight      = false  // loose/tight mode for lists
-  this.ddIndent   = -1 // indent of the current dd block (-1 if there isn't any)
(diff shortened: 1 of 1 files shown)


Iwagd-DHA (approved):

Iwagd-DHA (approved):

puzrin:
That's marked for the next version, and such trivial changes dn't need PR.

Issue #1144, opened 2025-12-16
HTML comment block incorrectly ended before `-->`


Jayman2000:
I’m trying to convert this chunk of Markdown (taken from [here](https://github.com/DescentDevelopers/Descent3/blob/0171d7b76a0c6d77908169af8ff3ea8d4d1cd930/BUILD.md#building---windows)) into HTML:

````markdown
1. **Install the prerequisite build tools.**

    Make sure that you have Git and Visual Studio 2022 with the “Desktop development with C++” workload and the “C++ MFC for latest v143 build tools (x86 & x64)” component. If you don’t already have those installed or you aren’t sure, then open an elevated Command Prompt and run:
    <!--
    The following code block specifies the full path to the Visual Studio Installer because the Visual Studio Installer doesn’t add itself to the user’s Path. The installer is guaranteed to be in a specific location on 64-bit systems [1]. The installer will be in a different location on 32-bit systems [2], but Visual Studio 2022 doesn’t support 32-bit systems [3] so we can ignore that detail.

    [1]: <https://learn.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio?view=vs-2022>
    [2]: <https://github.com/microsoft/vswhere/wiki#installing>
    [3]: <https://learn.microsoft.com/en-us/answers/questions/1689898/does-visual-studio-build-tools-2022-support-32-bit>
    -->
    ```batch
    winget install Git.Git Microsoft.VisualStudio.2022.Community

    "%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\setup.exe" modify^
        --passive^
        --channelId VisualStudio.17.Release^
        --productId Microsoft.VisualStudio.Product.Community^
        --add Microsoft.VisualStudio.Workload.NativeDesktop;includeRecommended^
        --add Microsoft.VisualStudio.Component.VC.ATLMFC
    ```
````

[...]

PR #1146, opened 2026-01-20
Fix HTML comment blocks in list items (issue #1144)

mihir0k:
## Summary
- Prevent HTML comment blocks (type 2) from terminating early on blank lines inside list items.
- Add a regression fixture for issue #1144.

## Details
The HTML block parser currently ends when it sees an outdented line. Inside list items, blank lines can be outdented but still belong to the list item, which causes `<!-- ... -->` blocks to end before `-->`. This change allows empty lines to remain inside HTML blocks when parsing list items, matching CommonMark’s type 2 rules.

## Testing
- npm test
- npm run benchmark-deps
- ./benchmark/benchmark.mjs


puzrin:
Sorry, I'm not comfortable with this change. In the reference CM parser, comment blocks do not depend on indent at all.

PR #1180, opened 2026-06-02
Don't strip spaces from code spans that consist only of spaces

sarathfrancis90:
### Problem

Code spans whose contents consist entirely of spaces have one space incorrectly stripped from each side.

```js
const md = require('markdown-it')();

md.renderInline('`   `');     // 3 spaces
// actual:   <code>  </code>   (only 2 spaces — one stripped from each side)
// expected: <code>   </code>  (3 spaces preserved)

md.renderInline('``    ``');  // 4 spaces
// actual:   <code>  </code>   (2 spaces)
// expected: <code>    </code> (4 spaces preserved)
```

This violates the CommonMark spec. The code-span normalization rules state:

> - If the resulting string both begins *and* ends with a space character, but **does not consist entirely of space characters**, a single space character is removed from the front and back.

The spec further makes this explicit with a dedicated example ("No stripping occurs if the code span contains only spaces"), and the reference implementation (commonmark.js) behaves accordingly.

### Root cause

In `lib/rules_inline/backticks.mjs`, the normalization used:

```js
.replace(/^ (.+) $/, '$1')
```

`(.+)` matches any one-or-more characters, including a run of spaces. So for input `` `   ` `` (3 spaces), the content `"   "` matches `^ (.+) $` with the capture group being the single middle space, stripping a space from each end. The regex never enforced the spec's "does not consist entirely of space characters" condition.

### Fix

Require the captured group to contain at least one non-space character, so all-space code spans are left untouched while normal content still has its single leading/trailing space removed:

```diff
-          .replace(/^ (.+) $/, '$1')
+          .replace(/^ (.*[^ ].*) $/, '$1')
```

This matches both the spec prose and the commonmark.js reference behavior. Content that has surrounding spaces and at least one non-space character (e.g. `` ` a ` `` → `<code>a</code>`) is unchanged.

### Tests

Added two regression fixtures to `test/fixtures/markdown-it/commonmark_extras.txt` covering the single- and multi-backtick all-space cases. They **fail** against the previous regex and **pass** with the fix (verified by reverting the source change and re-running).

Full suite is green locally:

- `eslint .` — clean
- commonmark spec tests — 652 / 652 pass
- markdown-it tests — 291 / 291 pass
- build test — 1 / 1 pass

No `dist/` artifacts are committed (gitignored; working tree clean after build). A quick micro-benchmark of the new regex vs. the old one over 5M iterations showed no measurable regression.

### DIFF
--- a/lib/rules_inline/backticks.mjs
+++ b/lib/rules_inline/backticks.mjs
@@ -41,7 +41,7 @@ export default function backtick (state, silent) {
         token.markup = marker
         token.content = state.src.slice(pos, matchStart)
           .replace(/\n/g, ' ')
-          .replace(/^ (.+) $/, '$1')
+          .replace(/^ (.*[^ ].*) $/, '$1')
       }
       state.pos = matchEnd
       return true
--- a/lib/rules_inline/backticks.mjs
+++ b/lib/rules_inline/backticks.mjs
@@ -39,9 +39,15 @@ export default function backtick (state, silent) {
       if (!silent) {
         const token = state.push('code_inline', 'code', 0)
         token.markup = marker
-        token.content = state.src.slice(pos, matchStart)
-          .replace(/\n/g, ' ')
-          .replace(/^ (.*[^ ].*) $/, '$1')
+        let content = state.src.slice(pos, matchStart).replace(/\n/g, ' ')
+
+        // One space is stripped from each side when both are present,
+        // unless the content is nothing but spaces.
+        if (/[^ ]/.test(content) && content[0] === ' ' && content[content.length - 1] === ' ') {
+          content = content.slice(1, -1)
+        }
+
+        token.content = content
       }
       state.pos = matchEnd
       return true
--- a/test/fixtures/markdown-it/commonmark_extras.txt
+++ b/test/fixtures/markdown-it/commonmark_extras.txt
@@ -869,16 +869,15 @@ foo
 <!bar baz>
 .
 
-Code span containing only spaces is not stripped (3 spaces)
+Code spans consisting only of spaces are not stripped
 .
 `   `
-.
-<p><code>   </code></p>
-.
 
-Code span containing only spaces is not stripped (multiple backticks, 4 spaces)
-.
+` `
+
 ``    ``
 .
+<p><code>   </code></p>
+<p><code> </code></p>
 <p><code>    </code></p>
 .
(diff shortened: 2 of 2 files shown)


puzrin:
I would rewrite this to `[^ ]` scan + conditional slices, to express the spec logic explicitly. Regexp was not cool initially, and after the update, it is even more difficult to understand.

PS. Also note for the next time, 2 links to markdiwn-it / cm dingus demos with examples are much better than long description, to understand the problem fast.

test/fixtures/markdown-it/commonmark_extras.txt
```diff
@@ -834,3 +834,17 @@ b
...
 .
+
+Code span containing only spaces is not stripped (3 spaces)
+.
+`   `
+.
+<p><code>   </code></p>
+.
+
+Code span containing only spaces is not stripped (multiple backticks, 4 spaces)
```
puzrin: I'm not sure why this is needed. Anyway, it's worth joining all trivial patterns t single block.

lib/rules_inline/backticks.mjs
```diff
@@ -41,7 +41,7 @@ export default function backtick (state, silent) {
         token.markup = marker
         token.content = state.src.slice(pos, matchStart)
           .replace(/\n/g, ' ')
-          .replace(/^ (.+) $/, '$1')
+          .replace(/^ (.*[^ ].*) $/, '$1')
       }
```
puzrin, line 51: Replace with a separate check for "all whitespaces", then slice() head/tail if needed. A single regexp is short, but not entirely readable.

Iwagd-DHA (approved):

Iwagd-DHA (approved):

Iwagd-DHA (approved):

sarathfrancis90:
Rewrote it as an explicit check — a `[^ ]` test for "not all spaces", then slice off head/tail. Tracks the spec wording much more directly than the regexp did.

Also folded the fixtures into one block. The multi-backtick case was not earning its own entry, so it is just another pattern there now, along with the single-space edge.

Apologies for the slow turnaround here. And noted on the dingus links — I will lead with those instead of a wall of text next time.

puzrin:
LGTM.

See dev branch, I already started big refactoring for v15 and will ask to rebase a bit later

sarathfrancis90:
Thanks. Saw the v15 work on dev — ping me when it's settled and I'll rebase onto it, no rush.

puzrin:
Rebase please.

jjrauch32-lab (commented):
ok

jjrauch32-lab (commented):
commit

PR #1182, opened 2026-06-17
fix: use ASCII-only whitespace in normalizeReference (CommonMark spec §6.6)

JSap0914:
## Bug

CommonMark spec §6.6 says link label normalization must strip leading/trailing **"spaces, tabs, and line endings"** and collapse consecutive internal **"spaces, tabs, and line endings"** to a single space.

The previous code used `.trim()` and `/\s+/g`, which both match Unicode whitespace characters (e.g. U+00A0 NO-BREAK SPACE) that the spec does not include in that set. This caused distinct reference labels that differ only by a non-ASCII whitespace character to be incorrectly normalised to the same key.

**Reproducer:**

```markdown
[foo\u00A0bar]: /url1
[foo bar]: /url2

[foo\u00A0bar]
[foo bar]
```

Before this fix both labels resolved to `/url1`. They should resolve to `/url1` and `/url2` respectively, because U+00A0 is not a "space, tab, or line ending" per the spec.

## Fix

In `lib/common/utils.mjs` → `normalizeReference`:

- Replace `.trim()` with `asciiTrim()` (already defined in the same file and used since #1074).
- Replace `/\s+/g` with `/[ \t\n\r]+/g` to restrict collapsing to the ASCII whitespace characters the spec names.

## Verification

```
npm run test:markdown-it   # 290 pass, 0 fail
npm run test:cmspec         # 0 fail
npm run lint                # 0 warnings
```

> AI-assisted: this PR was drafted with AI help.

### DIFF
--- a/lib/common/utils.mjs
+++ b/lib/common/utils.mjs
@@ -224,9 +224,12 @@ function isMdAsciiPunct (ch) {
 // Hepler to unify [reference labels].
 //
 function normalizeReference (str) {
-  // Trim and collapse whitespace
+  // CommonMark spec 6.6: strip leading/trailing spaces, tabs, and line endings;
+  // collapse consecutive internal spaces, tabs, and line endings to a single space.
+  // Non-ASCII whitespace (e.g. U+00A0 NBSP) is not in that set and must be
+  // preserved as a regular character.
   //
-  str = str.trim().replace(/\s+/g, ' ')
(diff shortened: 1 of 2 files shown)


puzrin:
I don't see a difference with CommonMark dingus, and the description looks like AI-generated garbage. If a problem exists, all you need is a trivial thing - provide links to dingus and markdown-it demos with examples of differences.

PR #1186, opened 2026-06-29
fix: handle non-BMP punctuation and symbols in smart quotes

mahirhir:
#1072 made `scanDelims` surrogate-pair aware so emphasis flanking handles astral punctuation/symbols. The smartquotes rule runs the same left/right flanking check to decide whether a `'`/`"` is opening, closing, or an apostrophe, but it still reads the neighbouring characters with `charCodeAt`. For a character outside the BMP that yields a lone surrogate, which counts as neither punctuation nor whitespace, so a quote next to it gets placed differently than next to the equivalent BMP character.

With `typographer: true`:

```
input    before     after
𝜵'x      𝜵’x        𝜵'x
𝜵"x"     𝜵"x"       𝜵“x”
"x"𝜵     "x"𝜵       “x”𝜵
```

`𝜵` is U+1D735 (a math symbol). The BMP nabla `∇` already produced the "after" output, so this only makes the astral case consistent with its BMP equivalent.

The change follows #1072: a small `getLastCharCode` helper for the two backward reads and `codePointAt` for the two forward reads. That helper is a copy of the one in `scanDelims`; happy to lift it into a shared util instead if you'd rather not duplicate it.

Added cases to `test/fixtures/markdown-it/smartquotes.txt`; the fixture fails on `master` without the lib change.

### DIFF
--- a/lib/rules_core/smartquotes.mjs
+++ b/lib/rules_core/smartquotes.mjs
@@ -7,6 +7,23 @@ const QUOTE_TEST_RE = /['"]/
 const QUOTE_RE = /['"]/g
 const APOSTROPHE = '\u2019' /* ’ */
 
+// Get the code point ending right before `pos`, taking surrogate pairs into
+// account. Mirrors the helper used by `scanDelims` (see #1072) so that the
+// flanking check here treats non-BMP punctuation/symbols the same way.
+function getLastCharCode (str, pos) {
+  /* eslint no-bitwise:0 */
+  // treat beginning of the line as a whitespace
+  if (pos <= 0) { return 0x20 }
+  const charCode = str.charCodeAt(pos - 1)
+  // not a low-surrogate code unit (is a BMP code point)
+  if ((charCode & 0xFC00) !== 0xDC00) { return charCode }
+
+  // undefined if out of range (e.g. an isolated low surrogate at index 0),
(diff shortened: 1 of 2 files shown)


puzrin:
Look at this:

https://markdown-it.github.io/#md3=%7B%22source%22%3A%22%5C%22x%5C%22qwe%20%5C%22x%5C%22%20qwe%22%2C%22defaults%22%3A%7B%22html%22%3Afalse%2C%22xhtmlOut%22%3Afalse%2C%22breaks%22%3Afalse%2C%22langPrefix%22%3A%22language-%22%2C%22linkify%22%3Atrue%2C%22typographer%22%3Atrue%2C%22_highlight%22%3Atrue%2C%22_strict%22%3Afalse%2C%22_view%22%3A%22src%22%7D%7D

Replace the nabla with any BMP character, and smartquoting will stay the same. Because it uses its own rule, NOT the emphasis. So, I see 2 problems here:

1. Examples are not as expected (BMP behavior is the same).
2. The result after the change is not as expected either.

I know, commanmark dingus works differently, but:

- Smartquiting behavior is package-specific (not specified)
- I don't like implementation in reference parsers. Changing a quote in the middle of a word looks ugly.

My recommendation: create an issue with real use cases to discuss first. That's simpler and requires less time.

PR #1193, opened 2026-07-15
demo: restrict permalink-loaded options to a presentation allow-list

superm1:
The demo page's `loadPermalink()` parses options out of the URL fragment and copies them into the live `defaults` object for any key that already exists with a matching primitive type. This allowed a crafted permalink to set arbitrary defaults, including `html`.

Enabling `html` turns on raw-HTML pass-through, and the rendered result is assigned to `.result-html`'s innerHTML, so a link of the form `#md3=...{defaults:{html:true},source:""}` would execute attacker-controlled script in the browser of whoever opened it.

This restricts the keys a permalink may set to a fixed presentation-only allow-list (`_view`, `_strict`, `langPrefix`, `breaks`, `linkify`, `typographer`, `xhtmlOut`). These control display/formatting only and cannot cause script execution. `html` is deliberately excluded. The existing same-type check and the legacy `_src` handling are preserved.

### DIFF
--- a/support/demo_template/index.mjs
+++ b/support/demo_template/index.mjs
@@ -380,6 +380,14 @@ function loadPermalink () {
 
   const opts = isObject(cfg.defaults) ? cfg.defaults : {}
 
+  // Presentation-only options that a permalink is allowed to set. These
+  // control display/formatting and cannot cause script execution. Notably
+  // `html` is excluded: it enables raw-HTML pass-through, whose output is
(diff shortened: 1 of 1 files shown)


puzrin:
IMHO garbage. The user can enable HTML directly from the permalink, and there is no reason to fight with it.

superm1:
Okay thanks.

PR #1194, opened 2026-07-15
cli: cap stdin buffering to avoid unbounded memory use

superm1:
When reading from stdin, the CLI buffered the entire stream into memory before rendering, with no upper bound on the amount of data collected. Piping a very large input could therefore grow the buffer without limit and exhaust available memory before any output was produced.

This tracks the cumulative size of the incoming chunks and aborts with a clear error once it exceeds a 32 MiB cap, stopping further reads from stdin. The file-based read path is unchanged.

### DIFF
--- a/bin/markdown-it.mjs
+++ b/bin/markdown-it.mjs
@@ -51,11 +51,25 @@ const options = cli.parse_args()
 function readFile (filename, encoding, callback) {
   if (options.file === '-') {
     // read from stdin
+    const MAX_INPUT_BYTES = 32 * 1024 * 1024
     const chunks = []
+    let size = 0
+    let done = false
 
-    process.stdin.on('data', function (chunk) { chunks.push(chunk) })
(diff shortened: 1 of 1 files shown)


puzrin:
Looks like AI slop. Can you explain the reason why this is really needed?

superm1:
Yeah; it was flagged from a security scanner using some of the frontier models.  The part that was flagged was that it's basically a way to trigger a DoS that could easily lead to an OOM.

The example the scanner gave was this leading to an OOM:

> yes | head -c 10G | markdown-it -

It was a relatively low score (1.8 CVSS) so I totally understand if you don't see a need to bound the CLI input, no hard feelings here.

puzrin:
Let's close this. CLI is not actually supported; it's for quick checks only. If the user has CLI access and wishes to shoot into the leg, that's doable anyway.

PR #1196, opened 2026-07-23
Remove obsolete Node.js 10 workaround in normalizeReference()

### LINKED ISSUE #1153: Remove legacy Node.js 10 workaround from `normalizeReference()`
The following workaround for a Node.js 10 bug should be removed:

https://github.com/markdown-it/markdown-it/blob/cdb89ad5bce40dce8791e55acf3a2ec373a22801/lib/common/utils.mjs#L231-L245
**Reasoning:**

- Node.js 10 reached End-of-Life in April 2021
- This bug was fixed in Node.js 12 (over 5 years ago)
- The workaround adds unnecessary complexity and confusion
- Modern Node.js versions handle this correctly
- `markdown-it` package is tested on Node.js 24

Hashim1999164:
## Summary
- Drops the Node.js 10-era `ẞ`/`Ṿ` special case in `normalizeReference()`.
- That runtime check is dead on current Node and browsers, so the branch never runs.

Fixes #1153

## Test plan
- [x] `npm test`

### DIFF
--- a/lib/common/utils.mjs
+++ b/lib/common/utils.mjs
@@ -228,17 +228,6 @@ function normalizeReference (str) {
   //
   str = str.trim().replace(/\s+/g, ' ')
 
-  // In node v10 'ẞ'.toLowerCase() === 'Ṿ', which is presumed to be a bug
-  // fixed in v12 (couldn't find any details).
-  //
-  // So treat this one as a special case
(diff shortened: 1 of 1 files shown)


puzrin:
Looks like AI garbage

PR #1200, opened 2026-08-14
fix: honor escaped characters in linkify autolinks

### LINKED ISSUE #1058: Since v13 `linkify` incorrectly parses links with escaped characters
The following script renders a Markdown string that contains a link with escaped characters:
```javascript
const markdownIt = require("markdown-it")({ linkify: true });
console.log(markdownIt.render(String.raw`See https://git\-scm.com/docs/gitignore\#\_pattern\_format.`));
```

With markdown-it@12, linkify parses the link correctly:
```html
<p>See <a href="https://git-scm.com/docs/gitignore#_pattern_format">https://git-scm.com/docs/gitignore#_pattern_format</a>.</p>
```

After upgrading to markdown-it@13 (or 14), linkify stops parsing the link at the first backslash:
```html
<p>See <a href="https://git">https://git</a>-scm.com/docs/gitignore#_pattern_format.</p>
```

arronKler:
## Description

With `linkify: true`, a CommonMark escape inside a `scheme://` URL stops the autolink at the backslash:

```
See https://git\-scm.com/docs/gitignore\#\_pattern\_format.
```

becomes `<a href="https://git">https://git</a>-scm.com/...` instead of one link to `https://git-scm.com/docs/gitignore#_pattern_format`.

The inline linkify rule matches `://` on raw source so emphasis inside a URL is not applied (v13). `\` is not a URL host character, so linkify-it ends the match there even though `\-` / `\#` / `\_` are just escaped punctuation.

## Fix

Unescape Markdown ASCII punctuation before `matchAtStart`, then advance `state.pos` by the corresponding source length (each `\` + punct counts as one character).

This does not change the v13 cases that are supposed to stay unlinked:

- `google\.com` (fuzzy / core linkify + `text_special`)
- `\https://...`, `https\://...`, `https:\//...` (escape still breaks the raw `://` trigger)

Wrapping with `<...>` remains the way to force an autolink when needed.

Fixes #1058

### DIFF
--- a/src/rules_inline/linkify.ts
+++ b/src/rules_inline/linkify.ts
@@ -1,9 +1,27 @@
 // Process links like https://example.org/
 
+import { isMdAsciiPunct, unescapeMd } from '../common/utils.ts'
 import type StateInline from './state_inline.ts'
 
 // RFC3986: scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
 const SCHEME_RE = /(?:^|[^a-z0-9.+-])([a-z][a-z0-9.+-]*)$/i
+
+// `\` + ASCII punct is one character after CommonMark unescape.
+function srcLengthForUnescaped (src: string, start: number, unescapedLength: number): number {
+  let i = start
+  let consumed = 0
+
+  while (consumed < unescapedLength && i < src.length) {
+    if (src.charCodeAt(i) === 0x5C/* \ */ && i + 1 < src.length && isMdAsciiPunct(src.charCodeAt(i + 1))) {
+      i += 2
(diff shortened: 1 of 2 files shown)


puzrin:
Closing.

The underlying issue isn't agreed to be a bug — escaping inside a URL hasn't been needed since v13, and my question in #1058 is still unanswered. Please agree on the scope in an issue first, as CONTRIBUTING asks; this is the second PR for the same unconfirmed problem (#1166).

Independently of that: unescaping the whole tail on every `://` is quadratic in document size (~13x slower than master on a paragraph with 5000 escaped links).

PR #1211, opened 2026-09-15
fix(smartquotes): keep converting quotes after the opener cap is hit

MFA-G:
The smartquotes opener cap added in cf27366 makes the rule give up on the rest of the paragraph, so ordinary quotes stop being converted.

### Problem

`process_inlines()` caps the opener stack at `MAX_OPENERS` (1000) to bound memory on malicious input. On reaching the cap it does:

```js
if (canOpen) {
  if (stack.length >= MAX_OPENERS) { return }
```

That `return` exits `process_inlines()` for the entire inline token stream — before the `replacements` map at the end of the function is applied. Two things break at once:

1. every replacement already collected for that paragraph is discarded, so pairs that *did* resolve revert to straight quotes;
2. nothing after the 1000th opener is looked at, so later quotes and apostrophes are never converted.

### Reproduction

markdown-it 15.0.2, `typographer: true`:

```js
const md = require('markdown-it')({ typographer: true })

const src = '"paired" ' + '"unmatched '.repeat(1100) + "isn't it"
const out = md.render(src)

out.includes('\u201cpaired\u201d')  //=> false, expected true
out.includes('isn\u2019t')          //=> false, expected true
```

Scanning the threshold with `'"quoted" ' + '"x '.repeat(n)`:

| n | `"quoted"` converted |
| --- | --- |
| 999 | yes |
| 1000 | yes |
| 1001 | **no** |
| 1100 | **no** |

A paragraph only needs 1000 unmatched opening quotes before a perfectly ordinary quote elsewhere in it silently stops being typographed. That is reachable in real prose — long quote-heavy transcripts, code-ish text, or pasted logs — not only in adversarial input.

### Fix

Skip pushing new openers past the cap instead of returning:

```js
if (stack.length < MAX_OPENERS) {
  stack.push({ ... })
  ...
}
```

Scanning continues, so replacements found before the cap survive and apostrophes after it are still converted. The memory bound is unchanged — the stack still never exceeds `MAX_OPENERS` entries; the only difference is that the openers beyond the cap are dropped rather than the whole paragraph.

### Validation

- `npm test` — 308 + 33 + 1 pass, 0 fail (lint, build, type-check, cmspec, markdown-it, build tests).
- Added a regression test in `test/markdown-it/misc.test.mjs`. Verified it **fails on the current code** (`not ok 4`) and passes with the change, so it pins the behaviour rather than merely describing it.
- The pathological suite still passes. Timing on the patterns that motivated the cap is unchanged and still linear:

  | pattern | time |
  | --- | --- |
  | `'"a '.repeat(100000) + "b' ".repeat(100000)` | 150 ms |
  | `"'a ".repeat(100000) + 'b" '.repeat(100000)` | 87 ms |
  | `'"a" '.repeat(100000)` | 122 ms |
  | `'"a '.repeat(200000)` (all openers, hits the cap) | 53 ms |

  Scaling on the all-openers pattern: 50k → 15 ms, 100k → 27 ms, 200k → 53 ms (4x input, ~3.5x time).


puzrin:
https://github.com/markdown-it/markdown-it/blob/master/CONTRIBUTING.md

Please create an issue first. All you need to provide is a real-world example where 1000 double quotes in a single paragraph is not abnormal.

